PR #72, opened 2014-09-22
Update 2014-09-03-deployment-to-elastic-beanstalk.md

andresriancho:


yanokwa:
If you have versioned buckets, where ever you have "s3:*ObjectAcl", you will also need "s3:*ObjectVersionAcl"

PR #168, opened 2015-01-26
Include test dependencies for go get

mlocher:
Based on feedback from a user. @ngauthier please take a look and let me know if this is fine!


ngauthier:
Flags have to be separate, like in my comment.

Also, we should not recommend -u -f. Just -t -v.

mlocher:
@ngauthier changed according to your feedback!

ngauthier:
LGTM

PR #184, opened 2015-02-17
Don't use a variable in the SFTP deployment

mlocher:


mlocher:
There is something major broken with the SFTP documentation, we need to setup a testing server and rework the whole section!

PR #188, opened 2015-03-03
Docs for pushing to a remote repository

mlocher:


Codebryo:
image

So I just booted the app and I really like it. For me its deffinitely GTM
:sheep: :shipit:

_posts/faq/2014-09-10-push-to-remote-repositories.md
```diff
@@ -19,10 +19,6 @@ error: Could not read c70d417b6c0f4536f9a288888e80955b78b1cbc7
 fatal: Failed to traverse parents of commit d9ff2fe276d15d485f9c71c309160d47cd4ccdfb
 ```
 
-This is due to the fact, that we perform a *shallow clone* of the repository to speed up build times.
+This is due to the fact, that we perform a *shallow clone* of the repository to speed up build times. We also only clone the specific branch which triggered the build, so if you want to push to another branch you need to fetch those first. The following two commands take care of this.
```
mcgaffin: I suggest you change:
> This is due to the fact, that we ...

to
 
> This is because we ...

_sass/docs/_gist.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * ------------------------------------------------------------------------------ *
+ * Styles specific for GitHub Gists
+ * ------------------------------------------------------------------------------ *
+ */
+.gist {
```
mcgaffin: Is the 8-space indentation intentional? 2-space is more common for css. It probably doesn't make a big difference, but it would be nice to have all codeship css everywhere indented the same way.

mlocher: Actually the indentation is done via tabs (and only 2 spaces wide on my screen), but I'll switch to spaces for scss files

PR #197, opened 2015-03-11
Enable more checks for the SCSS Linter

mlocher:


Codebryo:
gtm, and looks really nice. Scss linter looks great :shipit:

PR #206, opened 2015-03-24
Update build status documentation

mlocher:
@beanieboi could you take a look if this is fine?


mlocher:
We should update https://github.com/codeship/codeship-ruby#states with all new states as well!

PR #207, opened 2015-04-01
Feature/alert note blocks

Codebryo:
Possibility of using info and alert blocks


Codebryo:
Examples of usage
```
We had reports that using React.js with multiple Pipelines really sux.
{:.alert-block}

Did you know we now even support free Trials of ParallelCI for 14 Days ?
[Start Trial](#)
{:.info-block}
```

Result:
image

mlocher:
:+1: great, thanks @Codebryo

PR #210, opened 2015-04-15
Add information on installing bundler

mlocher:


beanieboi:
hmm, i thought rvm is installing bundler by default, isn't that the case anymore?

mlocher:
Apparently not for older ruby versions.

beanieboi:
boooooo :D

PR #211, opened 2015-04-17
Feature: Various mixpanel updates and fixes

clemens:
This PR adds the missing properties for episode 3 (see https://trello.com/c/TUOchySl/654-new-mixpanel-tracking-episode-3-extend-the-current-tracking-to-all-other-public-pages) and removes old code (see https://trello.com/c/N5GCC0Wy/708-new-mixpanel-tracking-episode-9-remove-old-events).

### DIFF
--- a/_includes/mixpanel.html
+++ b/_includes/mixpanel.html
@@ -1,5 +1,26 @@
 <!-- configure some basic information for mixpanel -->
-<script>
+<script>  
+  var query = window.location.search.substring(1);
+  var vars = query.split('&');
+  var params = {};
+  for(var i = 0; i < vars.length; i++) {
+    var pair = vars[i].split('=');
+    params[pair[0]] = pair[1];
+  }
+
+  var properties = {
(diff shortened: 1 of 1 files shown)


_includes/mixpanel.html
```diff
@@ -1,10 +1,35 @@
...
+    "NEW - First Ever UTM Term"          : params["utm_term"]
+  };
+
+  mixpanel.register_once(properties);
+  mixpanel.people.set_once(properties);
+
+  mixpanel.track("NEW - Page Viewed", {
+    "NEW - Current Page Title"            : document.title,
+    "NEW - Current Page URL"              : document.URL,
+    "NEW - Current Domain"                : "docs.codeship.com",
```
clemens: I briefly discussed this with @AlexTi: Even though this strictly speaking isn't the "current domain", I think it makes sense to pretend it is (a virtual domain, if you will) and split the results in Mixpanel.

PR #217, opened 2015-04-27
Should properly link to machine user documentation

chadfawcett:
This fixes an issue where the link to Machine Users Documentation was being treated as relative, resulting in a 404 on the Codeship documentation page.


mlocher:
Ups, thanks for fixing this!

PR #248, opened 2015-07-20
Add Codeship GPG key

mlocher:
key will be available at https://codeship.com/documentation/codeship.asc once this PR is merged.

/cc @flomotlik


flomotlik:
:shipit:

PR #249, opened 2015-07-24
Fixing broken link

hicksca:
http://learn.github.com/p/branching.html is no longer a valid link. Below was the closes how-to link on Github that I could find:
https://github.com/blog/1377-create-and-delete-branches


mlocher:
Thanks for the PR @hicksca, highly appreciated!

PR #263, opened 2015-09-03
Use the Codeship Docker infrastructure

mlocher:
# ToDo

* [x] Encrypt environment variables for the deployment
* [x] Deployment
* [x] Cleanup rebase
* [x] Wildcard deployments (i.e. `master` [branch] -> `documentation` [folder on S3], `staging/(.*)`-> `staging/$1`, `private/(.*)` -> `private/$1`)
* [x] Testing of the lifecycle policy (especially regarding the assumption that a redeploy reset the expiration dates)
* [x] `robots.txt` for the bucket
* [x] Special treatment for the `master` branch, to deploy to the `documentation` subfolder.
* [x] CNAME for the staging / private deployments (see codeship/dns#56)
* [x] Redirect S3 bucket root to https://codeship.com/documentation/
* [x] Fix the autoprefixer gem (done via switching to Gulp for auto-prefixing)
* [x] Cache the build container (currently using my personal Quay.io account)
* [ ] Change Mothership to use the new S3 bucket (see codeship/mothership#3086)
* [ ] Switch the links to the private Docker documentation on Mothership over to the new URL
* [ ] Authentication for `private` deployments.
  * We could use http://www.s3auth.com/ or try something else?
* [ ] Clean up documentation related buckets on S3 after switching over to Jet.

# Preview
See http://documentation.codeship.com/documentation/ for the latest deployment of the `master` branch. (Please note, that a new build isn't triggered for every push to this remote repository).

# Current issue

none


Dockerfile.aws
```diff
@@ -0,0 +1,9 @@
+FROM codeship/aws-deployment:latest
+MAINTAINER marko@codeship.com
```
bfosberry: maybe dev@codeship? I think we use that elsewhere

PR #266, opened 2015-09-11
Switch to the new AES key for the project

mlocher:
In the process of switching over to the Docker based infrastructure I had to create a new project on Codeship, which also created a new AES key. 

This PR re-encrypts the environment variables and Docker config files with the new key.


bfosberry:
:+1:

PR #267, opened 2015-09-12
Switch to Alpine Linux

mlocher:
Switch to a Alpine based build container (because the standard Ruby container used about 310m of space)

**Assumptions**: Dependencies for this project don't change that often

**Steps**
* Only install required dev packages and purge them after the gem / npm installation again
* Don't install gem documentation
* Switch back to a Node.js installation (as therubyracer is broken on Alpine)
* Run the Gulp steps on the same container

mlocher_documentation_ _quay_io

This brought build times without deployment down to about 2 minutes (if the cache can be used), deployments down to about 3 minutes (again with the cache).

Further improvements can be made by caching the AWS and CURL containers as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 # A sample Gemfile
 source "https://rubygems.org"
-ruby "2.2.3"
+ruby "2.2.2"
 
 gem 'jekyll', '~> 2.5.2'
 gem 'therubyracer', '~> 0.12.2'
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source "https://rubygems.org"
 ruby "2.2.2"
 
 gem 'jekyll', '~> 2.5.2'
-gem 'therubyracer', '~> 0.12.2'
 gem 'rouge', '= 1.9.0'
 gem 'sass', '~> 3.4.7'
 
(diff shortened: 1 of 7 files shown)


mlocher:
@bfosberry Do you mind taking a look? Thx!

beanieboi:
whoa, sweet!

beanieboi:
i think i will try this for lighthouse as well :D

Gemfile
```diff
@@ -1,9 +1,8 @@
 # A sample Gemfile
 source "https://rubygems.org"
-ruby "2.2.3"
+ruby "2.2.2"
 
 gem 'jekyll', '~> 2.5.2'
-gem 'therubyracer', '~> 0.12.2'
```
beanieboi: :heart_eyes:

Gemfile.lock
```diff
@@ -81,7 +76,3 @@ DEPENDENCIES
   rouge (= 1.9.0)
   sass (~> 3.4.7)
   scss_lint
-  therubyracer (~> 0.12.2)
-
-BUNDLED WITH
```
beanieboi: please update your bundler version and regenerate that file. the `BUNDLED WITH` should be at the end. it gets dropped if you use a very old version of bundler

mlocher: I was wondering why this was removed again, probably because I had to downgrade to Ruby 2.2.2 (latest version on Alpine). Thx for spotting this!

mlocher:
Final container size with npm dependencies is roughly 28m, and I might be able to purge even more packages!

mlocher:
Rebased on `master` to get the changes from #268 and #269

PR #273, opened 2015-09-15
Fail builds on `master` for forked repositories

mlocher:
Reason being, that e.g. our `documentation-private` fork, could overwrite the master documentation with outdated information.

Also fix the RegEx for running deployments.


flomotlik:
:shipit:

PR #279, opened 2015-10-01
Stabilize the builds

mlocher:
* Switch back to the official Ruby containers, using the `slim` image. (Ruby on Alpine was segfault'ing too often)
* Fix the AWS deployments by copying all files from the bin/ directory into the container for use during the deployment & ChangeLog generation
* Upgrade to Ruby 2.2.3 (again)
* Update Dependencies


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM ruby:2.2.3-slim
```
ngauthier: why slim and then install build essential? Non slim includes the gem building tools.

Dockerfile
```diff
@@ -6,32 +6,29 @@ WORKDIR /docs
...
-    git \
-    libffi-dev \
-    nodejs \
-    ruby \
-    ruby-dev \
-    ruby-io-console \
-    ruby-json && \
-  echo "gem: --no-rdoc --no-ri" > ${HOME}/.gemrc && \
-  gem install bundler && \
+RUN export DEBIAN_FRONTEND=noninteractive && \
```
ngauthier: I don't think you need the DEBIAN_FRONTEND any more, but I could be wrong

mlocher: I didn't see it in the `Dockerfile` for the base image, so I added it :)

Dockerfile
mlocher: Because with the standard image I'll get a ~300MB  container, with the SLIM variant I need to install (and purge) more stuff, but I only have a ~100MB container at the end.

Dockerfile
ngauthier: I mean I don't think you need this option. I think apt will be ok without. But I am not sure.

Dockerfile
ngauthier: But we don't deploy the container at the end right? We just s3 push stuff?

jschley:
lgtm

ngauthier:
Yep. :+1:. I think there are improvements to be made yet to the Dockerfile, but it's not critical.

Dockerfile
mlocher: Yeah, we only push to S3, but downloading / caching the container took a long time previously (that was the main reason I switched to Alpine)

mlocher:
Removed the `DEBIAN_FRONTEND` environment variable, let me know of the improvements, will create a second PR for those!

ngauthier:
It would be mostly to just simplify the dockerfile by using larger images and keeping packages. Favoring simplicity over build time.

mlocher:
Ah, for me it's the other way round, especially with the docs, I want the build to be as fast as possible, because the container is pretty simple anyways and won't get touched often.

ngauthier:
Cool. Sounds good!

PR #283, opened 2015-10-06
adding import project to org doc

BoVice:


mlocher:
Couple of things:

1. There is a "Administration" category, which imho is a better fit for this article. There's also already an article on [Organizations](https://codeship.com/documentation/administration/organizations/#adding-projects), which we could extend (or split into multiple articles)
2. Please use either the Codeship or Demoship organizations in the screenshot(s)
3. I think we can combine those three images into a single one, which has pointers for all three steps.

Imho, we should work on getting more and more articles from the FAQ category to the categories they should be organized within and not create more articles there. Right now, whenever there isn't already a very clear category we throw everything into FAQ :(

mlocher:
As the build failed, is this PR based on the latest changes from `master`?

BoVice:
:shipit:

PR #293, opened 2015-10-09
Add documentation for modular build environment

flomotlik:
We want customers to understand the build environment
with Docker is very modular and easy to extend. Decided to put
it into the services documentation and not into a separate article
as customers have to read the services docs and should not jump
over this specific text and simply not read the other article in the docs


_posts/docker/2015-05-25-services.md
```diff
@@ -14,6 +14,16 @@ categories:
 * include a table of contents
 {:toc}
 
+## Modular environments instead of one size fits all
+
+One of the issues that we set out to solve with our new infrastructure was providing a more modular and composable build environment for our customers. Why should the tools that are only necessary during your deployment be installed in the environment that your tests run in? Why can’t we build small containers, a plugin infrastructure if you want, and plug those specialized containers into the parts of the build where we need them?
```
AlexTi: composable build environment ~~for our customers~~. 

What do you think about removing the questions, but adding use cases examples instead.

_posts/docker/2015-05-25-services.md
```diff
@@ -14,6 +14,16 @@ categories:
...
+
+With Docker at the basis of our new infrastructure, this has now become possible. You can pull in any container you like to compose your build environment. These can be official Docker containers, provided by different companies like Codeship, or even your own pulled from a private registry. You can connect those containers to build your environment that can go from simple database support for your tests to matching exactly what you have in production in a service oriented infrastructure to run complex integration tests.
```
AlexTi: > provided by different companies like Codeship

Do we have a link to our Dockerhub repo?

_posts/docker/2015-05-25-services.md
```diff
@@ -14,6 +14,16 @@ categories:
...
+
+If you want to have a unified process and toolset to deploy or test applications across your company, you can even put those tools and scripts for deployment into a Docker container, push it into a private registry and use this container across your company’s builds. This makes a unified build environment across all your projects easy.
+
```
AlexTi: Can we add more use cases, examples here?

_posts/docker/2015-05-25-services.md
flomotlik: What kind of use cases do you have in mind? I think use case descriptions would probably need quite a bit more text to set them up properly.

And I think the questions are stronger rhetorically, as after those questions they ask themselves "Yeah why isn't this possible, I really want this now" Then we come in with "And here is how you can do it in Codeship" and in the back of their heads it should be "But I can't do this with my existing system and that sucks, so Codeship is clearly better".

We're creating a need that people didn't know they have potentially with the questions and then can deliver immediately.

_posts/docker/2015-05-25-services.md
flomotlik: Great idea, will add link to https://hub.docker.com/u/codeship

_posts/docker/2015-05-25-services.md
flomotlik: What kind of use cases did you have in mind?

AlexTi: At least the use cases you covered with your questions.
e.g.:
+ Deployment container
+ Container that only need to be used for certain branches, e.g. if you want to run integration tests only on master you can install all needed tools in there.

It is to give an idea of concrete examples.

flomotlik: Added a few examples

AlexTi:
lgtm

PR #307, opened 2015-11-16
Use two digit example numbering

mlocher:
see codeship/codeship-tool-examples#15


rheinwein:
Will merge https://github.com/codeship/codeship-tool-examples/pull/15 and then this one.

PR #315, opened 2015-12-16
adding warning about cached dependency conflicts

BoVice:


BoVice:
After discussing with @mlocher, we decided to include this in `common_errors` for when `npm install` fails.

mlocher:
talked to @BoVice about this, we decided to add a warning message directly to the builds, which show this behavior.

PR #322, opened 2016-01-25
Remove outdated link in documentation

flomotlik:


mlocher:
:shipit:

bfosberry:
We actually forked this to codeship/go-encrypt

flomotlik:
I'm not sure even then if we should keep that in the docs. Imho the advantage of having it there is rather low and we can keep the content more focused on features. @mlocher thoughts?

mlocher:
I agree, if we add technical details on the `jet` CLI we can add a link to the fork, I don't think this is required at the moment.

bfosberry:
:+1:

PR #326, opened 2016-02-04
How to start the Cassandra service

mlocher:
**Note**, this PR requires an update to the build VMs, which is not live yet. As such the `sudo` command mentioned will not work on the current build VMs.

(Future) public URL https://codeship.com/documentation/databases/cassandra/

Do not merge yet, relies on codeship/checkbot#837 :)


jschley:
lgtm!

PR #332, opened 2016-02-07
Add flag in docs to run the maven build in batch mode

ddebree:
To avoid logging messages when dependencies are being downloaded, the batch mode maven flag can be used. 
In the service definition file, the flag is already added, but in the steps file it was missing


mlocher:
Thanks for the PR @ddebree, highly appreciated :)

PR #333, opened 2016-02-08
Use system libraries for Nokogiri on Ruby 2.3

mlocher:
Explain how to install Nokogiri when using Ruby 2.3 using the system libraries and not the bundled versions.


rheinwein:
:+1: thank you!

PR #341, opened 2016-02-19
update CoC to the latest version

beanieboi:
there are couple new versions of the CoC. this updates to the latest version from http://contributor-covenant.org/.

it includes good examples and is more expressive in every way.


mlocher:
:shipit: thanks for the PR @beanieboi

PR #348, opened 2016-03-01
Fix error in Dockerfile example for Docker/PHP doc

jschley:


mlocher:
:shipit: once the tests are green

PR #349, opened 2016-03-02
Rework how to cache Python 3 dependencies

mlocher:
removing the old virtual environment caused the dependency cache to break beacuse of a missing folder.


BoVice:
LGTM

PR #352, opened 2016-03-03
Fix a typo in the Docker Ruby tutorial

mlocher:
The steps file mentioend a `specs` task, but the shell script defined the task as `spec`.


BoVice:
:shipit:

PR #360, opened 2016-03-14
Php edit

BoVice:
Moved xdebug snipped under **compose** as this is enabled by default specifically when running compose. It may run by default for other applications but this is the most popular use case for customers looking to disable it.


BoVice:
Closing without merging

PR #365, opened 2016-03-21
Docs for installing a custom PHPUnit version

mlocher:
Add information about upgrading the installed version of PHPUnit on the classic build VMs.


moritzplassnig:
:shipit:

PR #371, opened 2016-04-01
Embed discourse comments

krames:
This PR embeds discourse comments into our documentation site.

Along with this PR we will need to add `community.codeship.com` as our embeddable host.  This setting can be found under **Admin > Customize > Embedding** of our our discourse install. You can refer to this [post](https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963) for more information.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -5,6 +5,10 @@ permalink: /:categories/:title/
 tags_dir: tags
 ubuntu_version: trusty
 
+# Discourse Comments
+# See https://github.com/oblakeerickson/jekyll_discourse_comments
+discourse_url: "https://community.codeship.com/"
+
 # Defaults for posts
 defaults:
   -
--- a/_plugins/discourse_comments.rb
+++ b/_plugins/discourse_comments.rb
@@ -1,0 +1,50 @@
+# Discourse Comments for Jekyll
+#
+# Generates Discourse comments for posts
+#
+# Author: Blake Erickson, Justin Gordon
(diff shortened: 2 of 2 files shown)


krames:
@mlocher @BoVice I thought it might be nice to extend our community discussions out to our documentation site. If this is something we aren't interested in pursuing at this point please feel free to close this out.

krames:
Oh! I also wanted to let you know that if you try to run this change locally it will not render the comment section and you will see the following in the javascript console:

```
Refused to display 'https://community.codeship.com/embed/comments?embed_url=https%3A%2F%2Fwww.codeship.com%2Fdocumentation%2Fdatabases%2Fpostgresql%2F' in a frame because it set 'X-Frame-Options' to 'SAMEORIGIN'.
```

I believe this error should would once the documentation is running under https://codeship.com/documentation/.

mlocher:
To be honest, I don't really like the approach the plugin takes. Embedding the comments is as easy as adding a snippet of JS to the template we use for [individual posts](https://github.com/codeship/documentation/blob/master/_layouts/page.html), which wouldn't require us to monkey patch the string class and make this more explicit (and imo not more complicated that this approach).

We might need to create a separate layout for the posts only to not show comments on overview pages, but that would still be a very easy change.

krames:
@mlocher I liked the idea of this plugin because it was separate from the template. That being said, I think you have a better appreciation for how this should be implemented so I am just going to close this out.

PR #375, opened 2016-04-13
Don't use env variables in FTP/SFTP deploys

mlocher:
* Don't use the `$HOME` environment variable for (S)FTP based deployments and reference the home directory via `~/` instead.
* Use the same target directory for each of the examples (on the remote server).


bfosberry:
Lgtm

PR #378, opened 2016-04-13
Updating dependencies

mlocher:
Update to the most recent versions of the dependencies.

Also fix two warnings / errors from the SCSS linter that came up in the first build.


rheinwein:
green build == :+1:

PR #382, opened 2016-04-18
Update to Jekyll v3

mlocher:
Update to the latest available release of Jekyll v3.1.2

There are lots of deprecations with error messages similar to 

```
Deprecation: A call to '{{ post_url languages/2014-09-03-go }}' did not match a post using the new matching method of checking name (path-date-slug) equality. Please make sure that you change this tag to match the post's name exactly.
```

They are related to jekyll/jekyll#3926 and could be fixed by removing the subdirectory from the call to `post _url` (so `{{ post_url 2014-09-03-go }}` instead of `{{ post_url languages/2014-09-03-go }}` in the above example). 

According to https://jekyllrb.com/docs/upgrading/2-to-3/#my-categories-have-stopped-working, we'd also need to move our main categories to the root directory and create a `_posts` subdirectory in each of them with the posts.

Neither of the above two issues is interfering with builds atm though.

Staging deploy available at http://documentation.codeship.com/staging/jekyll3/


BoVice:
lgtm

PR #383, opened 2016-04-18
Switch to single column overview pages

mlocher:
Use a single column on category and tag pages as was already configured for Docker related articles.


BoVice:
lgtm

PR #384, opened 2016-04-18
Fix 3 column layout.

mlocher:
I have no idea why, but removing the `letter-spacing` brought back the third column. 

It would be great if @codebryo could take a look in a spare minute.


BoVice:
LGTM. Also noticed that the the content in the articles use up a little more whitespace, which looks much much much better 😄  Good catch @mlocher

PR #385, opened 2016-04-18
WIP Include folders from remote repositories

mlocher:
- Configuration is done via a JSON file
- Requires JQ to be installed
- Does not work on Docker based builds yet
- No automated tests at the moment.


mlocher:
Pushed this, so I can get it working with Docker based builds and get some feedback on the general approach.

PR #386, opened 2016-04-18
Make checksum files for Jet available for download.

mlocher:
Link to the `sha256sums` file we generate when deploying a new version of `jet` and add a paragraph on how to verify the checksums (for Linux / Unix based OSes).

* Changed the `curl` commands to use the filenames from S3 so they match the filenames in the checksum file.
* Update the default `jet` version in the configuration files to `1.6.0`, this will be overwritten during deployment by the actual current version and is only used in local testing
* Update the `log` shell functions in a minor way.

See http://documentation.codeship.com/staging/jet-checksums/docker/installation/ for a staging deployment.

Requires codeship/jet#183

### DIFF
--- a/_data/jet.yml
+++ b/_data/jet.yml
@@ -1,2 +1,2 @@
 base_url: https://s3.amazonaws.com/codeship-jet-releases
-version: "0.9.5"
+version: "1.6.0"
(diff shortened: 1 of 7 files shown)


mlocher:
Did a full rewrite, directly link to the `sha256` files on S3 as this simplifies the logic quite a bit and if we ever switch to a different download server we only need to update the URL (as long as we keep download archives and the checksum files in the same folder).

This requires a change to the `jet` deployment though, as the files on S3 currently include additional folder information that we need to get rid of first.

bfosberry:
LGTM, could you rebase?

mlocher:
Rebased onto `master`. Will merge once the build is green.

PR #398, opened 2016-05-06
Add overflow on code and pre tags

mlocher:


codebryo:
I was not aware this was not in place. But please be sure the selector is `pre, pre code` as pure inline code elements should not be scrolled but in the text flow.

mlocher:
I don't know when / where we lost that, it was definitely working previously. Also pushed a new commit with your feedback @codebryo

codebryo:
:shipit:

PR #400, opened 2016-05-06
Updates

mlocher:
* Update NPM based dependencies
* Don't install `gulp` globally, but symlink the local installation to `/usr/local/bin`
* Switch to Ruby 2.3.1


BoVice:
LGTM

PR #401, opened 2016-05-07
Check for Jekyll deprecations

mlocher:
Fail the build if the Jekyll build log contains deprecation messages.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,6 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 set -e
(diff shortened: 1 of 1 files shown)


mlocher:
See jekyll/jekyll#4873 for a fix for the root cause of the deprecations. For now we could only work around this on our end by removing the path from all calls to `{% post_url %}`

mlocher:
Rebased on the current master, since we now upgraded to Jekyll v3.3. in #501

mlocher:
Rebased on current `master` branch and force-pushed to `deprecations` branch.

PR #402, opened 2016-05-09
Run "jekyll doctor" during linting

mlocher:
`jekyll doctor` -- Search site and print specific deprecation warnings


rheinwein:
:shipit:

PR #403, opened 2016-05-09
Adding common problems to private base image

ethangj:
Adding a few common issues to the doc for troubleshooting purposes.

PR #405, opened 2016-05-10
Cache link update

ethangj:
Fixing anchor link (had defined it with base url tag incorrectly)


mlocher:
Let's wait for #403 to build and merge it first (as you pushed all changes from that PR as well) and only merge this one afterward.

Content :shipit:

PR #411, opened 2016-05-10
Tag clarification on steps and push docs

ethangj:
Adding the same clarification about tag limitations to both docs.


mlocher:
:shipit:

PR #413, opened 2016-05-12
Fix a bug with parameters to CMD

mlocher:
this resulted in Jekyll exiting with the following error message: `Error:  getaddrinfo: Name or service not known`


nunopato:
LGTM

PR #416, opened 2016-05-18
Clarify which chars are escaped when using the environment settings

### LINKED ISSUE #195: Alert users about unexpected behaviour when escaping characters.
I've just gone through a little hell here while trying to setup a custom deployment script.

I had to take advantage of environment variables and it turns out the webapp panel automatically escapes all ```;``` but not ```$``` so it was a bit confusing for a while untill I figured out.

Please add a disclosure about this behaviour inside the proper doc post, which I linked bellow, and save others from going through this situation. I not issuing a PR because I have little knowledge about the issue beyond the experience I've just described.

[Set Environment Variables](https://github.com/codeship/documentation/blob/0722406e11/_posts%2Fcontinuous-integration%2F2014-09-03-set-environment-variables.md)

mlocher:
Fixes #195

### DIFF
--- a/_posts/continuous-integration/2014-09-03-set-environment-variables.md
+++ b/_posts/continuous-integration/2014-09-03-set-environment-variables.md
@@ -16,8 +16,10 @@ On the ***Environment*** page of your project settings you can enter the variabl
 We will export these environment variables to the environment your build runs via a command similar to:
 
 ```bash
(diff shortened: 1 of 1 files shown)


ethangj:
lgtm

BoVice:
lgtm :shipit:

PR #420, opened 2016-05-19
Updating caching doc with credential callout and image/log swap

ethangj:


ethangj:
@mlocher let me know if you think this solves the credentials not being clear enough

@BoVice let me know if you think this should be more templatey and how i might change it to fit better

mlocher:
👍

PR #422, opened 2016-05-23
Rewriting volumes for clarity and context

ethangj:


BoVice:
lgtm

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -16,13 +16,20 @@ categories:
 To follow this tutorial on your own computer, please [install the `jet` CLI locally first]({{ site.baseurl }}{% post_url docker/jet/2015-05-25-installation %}).
 </div>
 
-## Using Docker volumes
+## What Are Docker Volumes?
+Volumes are a way to pass data between containers, as well as between steps in your CI/CD process. Volumes are directories that other containers can read data from once they have been mounted either inside of one of your containers or on the host infrastructure itself.
```
mlocher: imo the explanation of volumes is not very clear to me. I'd switch to something like

>... Volumes are directories that other containers can read and write data to. If you bind a volume to a host directory, you can also persist files (and their modifications) between individual steps in the CI/CD process.

Especially the second part of the above paragraph is still very rough, but you can't mount a volume on the host (technically you can, but, as it's already a directory on the host, you don't need to).

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -16,13 +16,20 @@ categories:
...
 
-Volumes can be used to connect containers in your build environment, and to share build artifacts between containers and steps in your pipeline.
+This means you can have containers that work together without needing to tightly couple them or build out any direct communication between them. This is a great way to keep your containers efficient, and volumes solve a lot of problems in Docker-based architecture.
```
mlocher: ... in _a_ Docker-based architecture.

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -16,13 +16,20 @@ categories:
...
 
-Please see the [example in the codeship-tool examples repository](https://github.com/codeship/codeship-tool-examples/tree/master/07.volumes) of how to use volumes.
+## Using Docker Volumes With Codeship
```
mlocher: Can we get rid of the header? It would move the two ways to use volumes on level higher in the toc and we can move or remove the one paragraph from this header.

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -16,13 +16,20 @@ categories:
...
+There are two primary ways to use Docker volumes in your CI/CD process with Codeship.
 
-To connect containers using volumes, first define a volume on one container in your codeship-service.yml:
+### Passing Data Between Containers
+You can mount volumes in your containers and allow other containers to access that volume *during the same step*. A step here refers to an individual command in your `codeship-steps.yml` file. ([Learn more about your Steps file here.]({{ site.baseurl }}/docker/steps/))
```
mlocher: > _during the same step_
that's not quite correct. Other containers will be able to access the volume anytime, but the volume (along with its parent container) will be reset for each step.

As for the clarification regarding steps, do you think we can remove the first sentence and leave only the link.

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -31,7 +38,7 @@ data:
     - /artifacts
 ```
 
-You can then use the `volumes_from` declaration to mount this volume on another container.
+Then, for your remaining services that require access to the volume, you will use the `volumes_from` directive in your other service definitions and specify which service you are reading the volume from (i.e. the one specified with your `volumes` directive.)
```
mlocher: ... you are _mounting_ the volume from. It's read/write not read only (at least by default).

_posts/docker/tutorials/2015-09-04-docker-volumes.md
mlocher: Also, I'd rewrite the link to "Learn more about the [Steps file]()." (I don't like full sentence links, but this is a very minor detail and totally up for discussion. Same for the other full sentence links below.)

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -40,11 +47,12 @@ myapp:
     - data
 ```
 
-## Persisting volumes between steps
+[We have a downloadable example of this set up here.](https://github.com/codeship/codeship-tool-examples/tree/master/07.volumes)
 
-By default, all steps have an independent set of containers. As such, referencing a volume common to multiple steps will not reference the same base folder. This means not only containers, but also volumes, are isolated between steps.
+### Passing Data Between Steps
+The second setup solves the problem of persisting data *between* steps in your `codeship-services.yml` file. Since every step runs on a separate set of containers, date produced in one step is not normally available to a later step - but by mounting a volume on the host infrastructure you can persist data throughout your entire build.
```
mlocher: I'd remove the _in your codeship-services.yml file_ from the first sentence. We don't persist data in the yml file :)

I'd also change _normally_ to _usually_, but maybe this only sounds "wrong" to me. 

...by mounting a volume on the host ~~infrastructure~~ you can...

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -53,6 +61,19 @@ data:
     - /tmp/artifacts:/artifacts
```
mlocher: I'm not 100% certain, `/tmp/` works for this. I usually use relative paths like `./site/`

@bfosberry could you take a look?

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -53,6 +61,19 @@ data:
...
 ```
 
-With this static volume mounting, every container mounting a volume, regardless of the step, will share the same base folder on the host. This allows build artifacts to be shared between steps of your build. For an example of this, check out the [deployment example](https://github.com/codeship/codeship-tool-examples/tree/master/08.deployment-container) showing how to build and share build artifacts between steps.
+On all other services that need to access this data, you can use either the `volumes_from` directive as explained in the previous example or simply provide the exact same `volumes` directive (with the host:container mapping) on all services that require access to your volume.
+
+[We have a downloadable example of this set up here](https://github.com/codeship/codeship-tool-examples/tree/master/08.deployment-container)
+
+## Common Use Cases
+There are several common problems that volumes solve, including:
```
mlocher: Volumes solve several common problems, including:

_posts/docker/tutorials/2015-09-04-docker-volumes.md
```diff
@@ -53,6 +61,19 @@ data:
...
+
+* Using a file in a shared volume as a "health check" to look for service availability or completion
+
+* Passing credentials between steps or services
+
+* Creating compiled assets and artifacts via one service to be deployed or pushed to an image registry from another service
+
+* Avoiding re-work between steps (such as not re-seeding your database multiple times.)
```
mlocher: That one is a bit dangerous. Because in most cases all DB modifications would be persisted, not only the seed data. So the user would need to make sure to reset the database after each step. Maybe we can find a different example?

mlocher:
I'd also add a link to the relative Docker documentation

* https://docs.docker.com/engine/userguide/containers/dockervolumes/
* Docker, https://docs.docker.com/engine/reference/builder/#volume
* Compose, https://docs.docker.com/compose/compose-file/#volumes-volume-driver

mlocher:
Love the rewrite, great stuff, a couple comments but 😍

ethangj:
@mlocher hopefully addressed, re-review?

mlocher:
:shipit:

PR #423, opened 2016-05-23
added meta data to stop comments from being indexed and followed

BoVice:
google crawlers are picking up the following error from Discourse's embedded comment widget:

```
https://community.codeship.com/embed/comments?embed_url=https%3A%2F%2Fcodeship.com%2Fdocumentation%2Ftroubleshooting%2Fbuilds-are-not-triggered%2F Failed to load resource: the server responded with a status of 403 (Forbidden)
```

After browsing meta.discourse.org and the rest of the internet, the issue lies on the permission of the `embed` folder. Since we are using a hosted version of discourse, we are not able to fix it ourselves. 

### The good
As a work around so the google crawlers doesn't pick this up, I added some meta data to not index or follow the URL in this PR.

### The bad
We may not be able to track traffic from `codeship.com/documentation` when someone clicks on the **Continue Discussion** button

Source for solution:
http://www.robotstxt.org/meta.html

### DIFF
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -1,6 +1,10 @@
+<head>
+  <meta name="robots" content="noindex, nofollow">
+</head>
+
 <div id='discourse-comments'></div>
 <script type="text/javascript">
-  DiscourseEmbed = { discourseUrl: 'https://community.codeship.com/',
+  DiscourseEmbed = { discourseUrl: 'https://community.codeship.com/embed',
                      discourseEmbedUrl: '{{ include.url }}' };
 
   (function() {
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -4,7 +4,7 @@
 
 <div id='discourse-comments'></div>
 <script type="text/javascript">
-  DiscourseEmbed = { discourseUrl: 'https://community.codeship.com/embed',
+  DiscourseEmbed = { discourseUrl: 'https://community.codeship.com/',
                      discourseEmbedUrl: '{{ include.url }}' };
 
   (function() {


ethangj:
lgtm

moritzplassnig:
LGTM

mlocher:
TL;DR, this is broken, I'm going to revert.

Long story (probably not short): you found a `meta` tag [1] to deal with robots. You correctly assessed that you can forbid robots to index a page (`noindex`) and have them not follow any links on that page (`nofollow`).

You also found out, that a `meta` tag, needs to be nested inside a `head` tag [2].

The problem is, a `head` tag is only allowed once per page, and it needs to be (according to spec) the first child element of the `html` tag. The partial you added the tag to is not rendered directly after the `html` tag, but another one is ([_includes/head.html](https://github.com/codeship/documentation/blob/master/_includes/head.html)). Which is problem number 1, the elements are placed in entirely the wrong position, and there are multiple `head` tags present now. (In fact, for some reason, there is a third `head` tag present, that I'm going to remove in a separate PR.)

Luckily, this (problem 1) might have saved us some trouble, because (hopefully) crawlers just ignored the additional `head` tag and thus the instructions in the nested `meta` tag. 

Taken from [1], emphasis mine

> robots, defining the behavior that cooperative crawlers should have **with the page**. 

A `meta` tag always applies to the page it appears one, which brings us to problem 2. You told all crawlers not to index any site that includes that partial (so each and every documentation article) and not to follow any links it contains.

I guess that's not what you wanted to achieve. That's why I'm reverting this PR (and opening a different one to remove the other additional `head` tag).

We also need to talk about PR reviews because they clearly failed in this instance.

* [1] https://developer.mozilla.org/en/docs/Web/HTML/Element/meta
* [2] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head

mlocher:
Separate PR to remove the second (or third) `head` tag is available at #425

PR #427, opened 2016-05-24
More specific instructions for robot account credential retrieval

rheinwein:
I think Quay may have changed their UI, so I updated our docs to reflect the new workflow.


BoVice:
lgtm

PR #428, opened 2016-05-24
fix to disallow embedded scripts from googlebot

BoVice:
Proposing another approach to the SEO issue. I created an external js file `comments.js`, and moved the embedded code for the discourse comment section there. This would allow me to explicitly disallow Googlebot from crawling anything in `comments.js` while still being able to call the javascript function and display the comment section as expected. 

@codebryo What do you think of this approach?

### DIFF
--- a/_includes/comments.html
+++ b/_includes/comments.html
@@ -1,11 +1,8 @@
-<div id='discourse-comments'></div>
+<div id='discourse-comments'>
 <script type="text/javascript">
-  DiscourseEmbed = { discourseUrl: 'https://community.codeship.com/',
-                     discourseEmbedUrl: '{{ include.url }}' };
-
-  (function() {
-    var d = document.createElement('script'); d.type = 'text/javascript'; d.async = true;
-    d.src = DiscourseEmbed.discourseUrl + 'javascripts/embed.js';
-    (document.getElementsByTagName('head')[0] || document.getElementsByTagName('body')[0]).appendChild(d);
-  })();
+  // displays iframe after DOM loads
+  $(document).ready(function(){
+    comments();
+  });
 </script>
+</div>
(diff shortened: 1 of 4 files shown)


mlocher:
I'm 👎  on this approach, here's why:

1. There already is a [robots.txt](https://github.com/codeship/documentation/blob/master/_robots.txt), that get's deployed to the S3 buckets root folder (whereas the documentation itself get's deployed to a subfolder depending on the branch name). See https://github.com/codeship/documentation/blob/master/codeship-steps.yml#L49-L51 for the deployment steps.

2. We're already including jQuery at https://github.com/codeship/documentation/blob/6381a1485f92d72b2706c9331de867319e949ec7/_includes/footer.html#L41

3. If I interpret https://support.google.com/webmasters/answer/6062608?hl=en correctly, that would still not hide the embedded comments from the crawler, it would only disallow crawling if the crawler came to the site via a direct link. (But the Google crawler runs Javascript, so imo it would load the page, load the external JS file, run the JS, try to access the iframe URL and trigger the error)

  > You should not use robots.txt as a means to hide your web pages from Google Search results. This is because other pages might point to your page, and your page could get indexed that way, avoiding the robots.txt file. 

4. I looked through the Discourse logs, here's a link to one of the error messages that trigger this (only accessible with the correct permissions on our forum): https://community.codeship.com/logs/show/9e7e3ff4cf9af9609ad8534a0769de9d There is an actual error ocurring, which triggers the 403 status code. We should fix the error instead of hiding it.

I asked @manualwise to provide me with access to the Google Webmaster tools for https://codeship.com. I'd like to check how the Google Crawler sees a page like https://codeship.com/documentation/troubleshooting/builds-are-not-triggered/. When looking at the errors for the `/embed/` URLs I see a couple referrers, mostly https://codeship.com but other URLs (like https://codeship.io, or HTTP versions) as well. 

Reading through documentation, the 403 error can also be caused if the referrer doesn't match an allowed site. So maybe we need to fix our documentation hosting and adapt (some of) the redirects.

BoVice:
Closing until we agree on a solution

PR #429, opened 2016-05-31
Remove CI specific dockerfile

mlocher:
Re-use the main Dockerfile on CI as well. (And remove the unused `curl` service.)


ethangj:
lgtm (couldnt run locally because of setup issues though)

mlocher:
I'll wait with merging until we can rebuild https://github.com/codeship-library/aws-deployment. The current container is using an outdated version of the AWS CLI, which fails the updated AWS lifecycle configuration 😢

PR #433, opened 2016-06-06
Call Slack notification script via bash

mlocher:


ethangj:
lgtm but not an expert on this

mlocher:
Thanks, Ethan, this is merely a bug fix, the `sh` shell on the container had problems with certain variables, mainly `${CI_COMMIT_ID:0:10}`, which only uses the first ten chars of a variable.

`bash` expands the variable correctly, so we're switching :)

PR #438, opened 2016-06-08
short aws cli install doc

BoVice:
Adding a short doc with information on how to install the AWS CLI on Codeship's classic infrastructure. Will add additional details in the future.


mlocher:
What do you think about renaming the article to _Deploying to AWS_ (or something similar) 

The have a generic section about using the AWS CLI and point to the following integrated deployments as well?

* https://github.com/codeship/documentation/blob/master/_posts/continuous-deployment/2014-09-03-deployment-to-elastic-beanstalk.md
* https://github.com/codeship/documentation/blob/master/_posts/continuous-deployment/2014-11-10-deployment-to-aws-codedeploy.md
* https://github.com/codeship/documentation/blob/master/_posts/continuous-deployment/2015-06-16-deployment-to-aws-lambda.md

mlocher:
If you don't want to rework the article, I'm :shipit: on the current version as well :)

BoVice:
I was thinking the same thing to move all AWS articles into a single doc, but wanted to have something out quick for now as integrating all other AWS docs into one would be a rather large task.

BoVice:
Will definitely congregate all those docs sooner than later though 😄

mlocher:
I wouldn't move them all to the same doc, because that would be huge. But have an overview article and then more indepth ones for the integrrated deployments.

BoVice:
👍  Let's discuss the details offline. GTM otherwise?

mlocher:
> If you don't want to extend the article, I'm :shipit: on the current version as well :)
Absolutely

PR #448, opened 2016-07-13
Generator tutorial -- warning about remote Docker hosts

rheinwein:
Some auth generator services require access to the local filesystem in order to write the generated auth file. If you attempt to run any of these services using a remote Docker host (AWS, Digital Ocean) it will fail because it doesn't have access to write to your local host.


mlocher:
:shipit:

PR #449, opened 2016-07-13
Update configuration to new caching mechanism

mlocher:
* Remove credentials for Quay.io
* Remove the special `ci` service and replace it with the `documentation` service (which was already identical, but referencing the Docker Hub image name)
* Remove references to the `encrypted_dockerfg_path` from all steps but the `push` step


rheinwein:
:shipit:

PR #451, opened 2016-07-20
Don't pass --prefer-source for caching reasons

mlocher:
`--prefer-source` doesn't work well in combination with the dependency cache and forces a re-install of (most) dependencies.


BoVice:
LGTM

ethangj:
lgtm

PR #452, opened 2016-07-20
Dev

BoVice:


.gitignore
```diff
@@ -11,3 +11,4 @@ node_modules
 # OS & editors
 .DS_Store
 *.codekit
+codeship.aes
```
mlocher: `codeship.aes` is already included on line 7 of `.gitignore`

PR #455, opened 2016-07-26
Get more debugging information on Slack notifications

mlocher:
* Print the payload as a log message
* Switch curl from silent to verbose mode


rheinwein:
💯

PR #456, opened 2016-07-29
Fixing broken link in cache tutorial

rheinwein:
image


mlocher:
thanks for the fix, :shipit:

PR #459, opened 2016-08-04
Deploy the master branch to the bucket root too

mlocher:
For the `master` branch, generate the site twice, once at the root of the destination
folder and once in the `/documentation` subfolder.

This is a preparatory step for moving the documentation from
http://codeship.com/documentation to http://documentation.codeship.com. In a further PR
we will need to update the `robots.txt` to allow robots access directly via the bucket and
the S3 website configuration to redirect switch redirects to the new location.


BoVice:
lgtm

PR #466, opened 2016-08-22
update postgres env variables

beanieboi:
we export both versions `PG_USER`, `PGUSER` and `PG_PASSWORD`, `PGPASSWORD`.
this updates the docs to reference the official version going forward

https://www.postgresql.org/docs/8.4/static/libpq-envars.html


mlocher:
:shipit:

PR #478, opened 2016-09-07
Work around a CSS image with links in headings

mlocher:
There is a positioning issue with links in headings, to work around I switched to a nested list for now.

See http://d.pr/i/17Wu6 for a screenshot.


alextacho:
lgmt

PR #480, opened 2016-09-15
Adding docs for excluding a step

rheinwein:
The base functionality is in Jet, but the UI updates have not been pushed to Mothership.

https://www.pivotaltracker.com/story/show/106991536

### DIFF
--- a/_docker/getting-started/steps.md
+++ b/_docker/getting-started/steps.md
@@ -43,23 +43,26 @@ Steps are specified in a list, such as:
   command: echo bar
 ```
 
-Steps are executed in the order they are specified. All steps share two directives:
+Steps are executed in the order they are specified. All steps share these directives:
 
 * `name` specifies the name of the step, outputted in the logs. This must be unique for all steps. **This field is optional** and will be auto-generated if not specified.
-* `tag` this says to only execute this step for the given "tag". Within Codeship this will be the branch or tag name. You can specify this locally via the `--tag` flag. This field is optional.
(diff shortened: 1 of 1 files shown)


_posts/docker/2015-05-25-steps.md
```diff
@@ -42,23 +42,27 @@ Steps are specified in a list, such as:
   command: echo bar
 ```
 
-Steps are executed in the order they are specified. All steps share two directives:
+Steps are executed in the order they are specified. All steps share these directives:
 
 * `name` specifies the name of the step, outputted in the logs. This must be unique for all steps. **This field is optional** and will be auto-generated if not specified.
-* `tag` this says to only execute this step for the given "tag". Within Codeship this will be the branch or tag name. You can specify this locally via the `--tag` flag. This field is optional.
+* `tag` this field will cause the step to be executed only when the build tag matches the `tag` value. During a Codeship build, the build tag is either the branch name or tag associated with the commit. If you are using this feature locally, you can pass in `jet steps -
+   -tag TAG_NAME` to create a build tag. This field is optional.
```
mlocher: Is the line break on purpose?

rheinwein: Nope :) fixed

rheinwein:
@mlocher I updated the docs to note that `exclude` takes precedence, and added a test case for that in the Jet PR. I intentionally set it up like this because I think it's safer to skip a step than to run it when it wasn't intended to be run. Alternatively we could error the build if both are present, but I'm not sure that's necessary.

bfosberry:
@rheinwein could you take a look at this tomorrow, is this merge-ready?

rheinwein:
@bfosberry see updated description

PR #482, opened 2016-09-19
Redirect "docker/getting-started" to "docker-guide/getting-started"

mlocher:
We missed a redirect from the old _Getting Started_ guide to the new one. Google still links to the old URL, so let's redirect it :)

After this is merged https://documentation.codeship.com/docker/getting-started/ should redirect to https://documentation.codeship.com/docker-guide/getting-started/


BoVice:
lgtm

PR #483, opened 2016-09-20
Feature/update header and footer

codebryo:
Hello there,

this PR adds the latest header and footer style from our marketing pages to the documentation so it's looking alike.

To note is: The ribbon to contribute was moved down below the header image for the sake that it otherwise collides with the signup button.

image

image

Thanks,
Roman


codebryo:
Updated the linting. I ignore the bulk of precoded CSS as it's the foundation, but linted the rest of the file accordingly. I also enabled the linters after the block again.

PR #484, opened 2016-09-20
Add info message regarding API and the Docker platform

mlocher:
Also, fix a bug with the article not showing up on the index pages, because of a wrongly capitalized category name


ethangj:
lgtm!

PR #492, opened 2016-10-03
Fix typos

thoughtkettle:


_sass/rig/_functions.scss
```diff
@@ -8,7 +8,7 @@
 */
 
 
-// This Function let's you assign multiple Selectors to a list with :pseudo-selectors if needed
+// This Function lets you assign multiple Selectors to a list with :pseudo-selectors if needed
```
mlocher: FYI, those files are handled by @codebryo and pulled in from an external repository regularly. So it's very likely this error will make it back in the future

mlocher:
Not sure, what the build failure was about, I restarted the build and it's all green now. Will look into internal logs to see what went wrong with the original build :(

PR #493, opened 2016-10-04
Fix Dockerfile - missing curl

thoughtkettle:


Dockerfile
```diff
@@ -14,6 +14,7 @@ EXPOSE 4000
 RUN \
 	apt-get update && \
 	apt-get install -y --no-install-recommends \
+		curl \
 		apt-transport-https && \
```
rheinwein: ugh, thanks. they took `curl` out of `2.3.1-slim` a couple weeks ago, so fresh builds of the image run into this issue. ⭐️

mlocher: nitpicking here, but could we sort those packages alphabetically? (I guess there's a reason my girlfriend sometimes calls me Monk) 😄

PR #496, opened 2016-10-10
Fixed link

jp-co:
Noticed this link wasn't actually linked. Also improved case consistency of "CLI".


mlocher:
@jp-co thanks for taking the time to fix this :smile:

PR #498, opened 2016-10-13
Fixing inconsistent padding top and bottom on `tag` elements

danielcgold:
What is currently is
screen shot 2016-10-13 at 4 28 04 pm

With my change
screen shot 2016-10-13 at 4 27 54 pm


mlocher:
Thanks for taking the time to make this change :)

@codebryo could you take a look and comment/merge?

PR #515, opened 2016-10-31
Update FAQ to account for move to AWS VPC

GeorgeMac:
We're moving in to a VPC in AWS.

This means that we our networking is going to change. Our instances will be situated in a new Security group within our VPC. 

I would hold off merging this, until we are ready to release the VPC changes :+1:


GeorgeMac:
I would hold off merging this, until we are ready to release the VPC changes 👍

PR #523, opened 2016-11-09
Revert "Fix vw caused rounding error"

codebryo:
Reverts codeship/documentation#522


thoughtkettle:
Reverted because sidebar UI issues suddenly with border and text being bad.

PR #546, opened 2016-11-16
Hide tags on the documentation pages

mlocher:
Tags are still included in meta tags and used by Swiftype, but don't show up for the user.

Reason being that Jekyll doesn't support tags for arbitrary collections and we can't generate tag overview pages at the moment.


mlocher:
I ran a local version to make sure, thanks for the review!

PR #555, opened 2016-11-21
S3 Redirects

mlocher:
For each of the pages generated by the `jekyll-redirect-from` from gem, create a website redirect as documented on https://aws.amazon.com/blogs/aws/amazon-s3-support-for-website-redirects/

The script ins this PR looks through the generated site searching for the meta refresh tag, gets the target URL and then pushes the same file to AWS S3 with the added `--website-redirect-location` option set to the target URL.

That option will trigger a 301 permanent redirect on CloudFront.

Staging build available at https://app.codeship.com/projects/102044/builds/9dc53ef9-cb78-4707-aed3-84a9c014b811 and deployed to https://documentation.codeship.com/staging/s3-redirects-take1/.

See e.g. https://documentation.codeship.com/staging/s3-redirects-take1/databases/cassandra/  which redirects (via HTTP) to https://documentation.codeship.com/staging/s3-redirects-take1/classic/getting-started/cassandra/

### DIFF
--- a/bin/s3.sh
+++ b/bin/s3.sh
@@ -26,7 +26,7 @@ case "$action" in
 		;;
 	'configure_redirects')
 		if [ "${CI_BRANCH}" = "master" ]; then
-			cd /site/_site || exit 1
+			cd /site/master/ || exit 1
 		else
 			cd /site/ || exit 1
 		fi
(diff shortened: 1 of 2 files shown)


bin/s3.sh
```diff
@@ -24,6 +24,26 @@ case "$action" in
 		config_file=${1:?'You need to pass an action!'} && shift
 		aws s3api put-bucket-lifecycle --bucket "${AWS_S3_BUCKET}" --lifecycle-configuration "file://${config_file}"
 		;;
+	'configure_redirects')
+		if [ "${CI_BRANCH}" = "master" ]; then
+			cd /site/_site || exit 1
+		else
+			cd /site/ || exit 1
```
bfosberry: I feel like you dont need to || exit 1 along with set -e, is there a reason you're using that as a backup?

mlocher: `shellcheck` complains about this (at least the version I have installed in my editor) :)

PR #571, opened 2016-12-12
Create a project article

thoughtkettle:
* merged some one-liner articles or articles with little content that had the same interface window or are generally related

* changed Classic welcome page to reflect the general create project article

* added/moved screenshots

* created create a project article that is relevant for both Classic and Docker


_general/projects/getting-started.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Getting Started With Projects
+layout: page
+tags:
+  - administration
+  - project
+category: Projects
+redirect_from:
```
mlocher: There are 4 articles getting combined into one, each with a redirect. Aren't we missing some redirects? (For all articles except "Delete a project" the current URLs are missing)

PR #574, opened 2016-12-19
Article for setting up a different time zone

BoVice:
This question is still brought up enough times in support requests that an article should be available to address how to set this up.


mlocher:
You could switch the link to the man page to use the `man_url` helper (see https://github.com/codeship/documentation#man-pages) but not a necessary change.

PR #576, opened 2016-12-20
Update timezone doc

joesiewert:
Fixed a couple links and the command syntax.


BoVice:
doy, I fudged up the links 😔 sending another fix

BoVice:
I just pushed a commit to add the proper url from man8. The `man_url` is configured for man1 pages. For this particular case, the doc for from man1 is a description of a different process that requires root-level access.

PR #582, opened 2016-12-21
codeship-services.yml updates to reflect V2 support

rheinwein:
Hey all,

A couple changes to codeship-services.yml:

* Users can now use V2 files with Codeship. We will ignore the `version` header and attempt to parse everything nested under the `services` key as a normal Codeship service
* Our build directive now supports `context`, which is the same as `path`. 

None of these changes introduce a breaking change for customers. In an attempt to unify our syntax, I am marking both `dockerfile_path` and `path` as deprecated, as those were custom Codeship fields in the build directive that have since been replaced by a Docker-specified key. We should encourage users to just use the Docker Compose build syntax instead of the custom Codeship syntax; it's easier.

In general, we want to focus less on "V2 Support" or "Docker Compose parity" and emphasize that the files are similar but not identical, and make it easier for customers using V2 files to get started with Codeship quickly.

https://www.pivotaltracker.com/story/show/136389287


_docker/getting-started/services.md
```diff
@@ -16,27 +16,52 @@ redirect_from:
...
 
-* Set up different environments for running your unit, integration or acceptance tests to keep containers small and startup times fast.
-* Build specialized deployment containers to unify deployment across your company
-* Use any Docker container available on an image registry for your builds
-* Easily switch between different versions of databases or build your own container for a special database
-* Have full control over every piece of Software in the build
-
-**Note**: There are a few Docker Compose features not yet supported by `codeship-services.yml`. We will eventually be at full Docker Compose parity, but until then you can use most features of Docker Compose in your `codeship-services.yml` file with little to no modification. For a list of non-supporter Docker Compose features, see below in this article.
+* Set up different environments for running your unit, integration, or acceptance tests
+* Have full control over your dependencies, versions, and every piece of software used in the build
```
mlocher: Should we mention that we don't offer control over the version of Docker used for the build?

rheinwein: Yeah, great point. Will update.

_pro/getting-started/services.md
```diff
@@ -16,27 +16,54 @@ redirect_from:
...
+    build: .
+    environment:
+      ENV: my-var
+  data:
+    image: busybox
+    volumes:
+      - ./tmp/data:/data
+``` 
+
+If include a `version`, Codeship will ignore the value, as the features supported by Codeship are version independent.
```
mlocher: missing a _you_ I guess?

pro/index.md
```diff
@@ -5,13 +5,12 @@ redirect_from:
...
+Welcome to Codeship Pro, the most customizable way to run your tests on Codeship. Easily mirror your development, test and production environments with full parity. The underlying build infrastructure, based on Docker, allows for customized definition of each service and dependency you require.
 
-<div class="info-block">
-Codeship Pro is available to customers on a invite only basis at the moment. Please see the [Docker feature page](http://pages.codeship.com/docker) for information and to request access.
+Your Codeship Pro builds run on infrastucture equipped with version {{ site.data.docker.version }} of the Docker Engine.
 
-**Note**
-- Codeship Pro supports git-based repositories on GitHub, GitLab, and BitBucket. **Mercurial based repositories on BitBucket are not currently supported.**
+<div class="info-block">
+**Note** Codeship Pro supports git-based repositories on GitHub, GitLab, and BitBbcket. Mercurial-based repositories on Bitbucket are not currently supported.**
```
mlocher: missing a _u_ on the first mention of Bitbucket (also lowercase second _b_)

PR #588, opened 2017-01-03
fixing broken link for 'existing script'

BoVice:


BoVice:
woops, good catch @joesiewert

PR #593, opened 2017-01-05
Add redirect for bit.ly/codeship-jet-tool

joesiewert:
Tested this locally and it should do the trick to get that bit.ly link working again.

### DIFF
--- a/_pro/getting-started/installation.md
+++ b/_pro/getting-started/installation.md
@@ -11,6 +11,7 @@ tags:
 category: Getting Started
 redirect_from:
   - /docker/installation/
+  - /documentation/docker/installation/
 ---
 
 <div class="info-block">


rheinwein:
Hmm, I'm not sure this will fix it -- The `bit.ly` URL is pointing to `/documentation/docker/installation` on codeship.com, not documentation.codeship.com. We already redirect from `/codeship.com/documentation/docker/installation/` to
`documentation.codeship.com/docker/installation/`, and then it breaks when we redirect to the `pro` directory. Am I missing something?

We kind of have a mess of redirects here:
http://getlinkinfo.com/info?link=bit.ly%2Fcodeship-jet-tool&x=36&y=16

joesiewert:
I was able to see it fail and then work locally after adding that redirect.  Worth a try I think.

rheinwein:
Cool, just wanted to make sure I understood how the redirects work -- ship it!

PR #600, opened 2017-01-09
Parallel steps clarification

ethangj:


mlocher:
👍 from me, @rheinwein what do you think?

PR #604, opened 2017-01-11
Adding pg_dump info

BoVice:
Adding some info for dealing with `pg_dump` errors in your build on Codeship Basic

### DIFF
--- a/_basic/databases/postgresql.md
+++ b/_basic/databases/postgresql.md
@@ -75,6 +75,20 @@ Similar to the other versions, you need to work around our auto-configuration fo
 sed -i "s|5432|5436|" "config/database.yml"
 ```
 
+### pg_dump
+Occasionally, you may experience a version mismatch with the PostgreSQL version you have configured.
+
+```shell
+pg_dump: server version: $YOUR_VERSION; pg_dump version: 9.2.19 
+pg_dump: aborting because of server version mismatch
+```
+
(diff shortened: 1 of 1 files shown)


_basic/databases/postgresql.md
```diff
@@ -75,6 +75,20 @@ Similar to the other versions, you need to work around our auto-configuration fo
 sed -i "s|5432|5436|" "config/database.yml"
 ```
 
+### pg_dump
+Occasionally, you may experience a version mismatch with the PostgreSQL version you have configured.
```
mlocher: I would mention, that this will happen if you use a non-default PostgreSQL version (so any version other than 9.2 at the moment).

BoVice: I think the wording already makes it pretty clear. It is already implied that you are not using the default version if you have already configured the pg version.

> ### 9.2
> 
> The **default version** of PostgreSQL on Codeship is **9.2**, which runs on the default port of `5432`. No additional configuration is required to use version 9.2.
> 

I'd like to limit redundancy and any further confusion. @joesiewert @ethangj what are your thoughts?

ethangj: I always like more info over less, but agree with @BoVice that this is pretty clear as is.

joesiewert: 👍

PR #608, opened 2017-01-13
adding quick fix around unexpected a:hover

BoVice:
Not ideal, but is a quick workaround. May inspect further if I have some downtime.


rheinwein:
I'd also be fine with this not being a header, and just being **bold** instead (since I caused the bug 🙈)

BoVice:
I'll do that then @rheinwein

PR #609, opened 2017-01-14
Clarifying tag language

ethangj:


ethangj:
@mlocher Laura says tag does still match. I will play with this a bit more. Can close this for now.

PR #612, opened 2017-01-24
Jenkins Migration Guide In Docs

ethangj:


ethangj:
@mlocher created a new dir for guides as we'll be moving some others (kubernetes) in here over time. Not sure what other config jekyll needs for this to work.

PR #614, opened 2017-01-25
updating to clarify build restarts

dennisnewel:
and to add source for codeship-build-restart script


mlocher:
@dennisnewel What do you think about moving the script itself to https://github.com/codeship/scripts (under the `utilities` folder)?

dennisnewel:
@mlocher we could do that too. would make it easier to copy. should we then completely remove the script from the documentation site?

mlocher:
@dennisnewel yeah, I'd just include a link to the script on the other repository. We do that with a couple other useful scripts as well (see e.g. https://github.com/codeship/documentation/blob/master/_basic/languages-frameworks/python.md#additional-versions)

dennisnewel:
ok. thanks @mlocher i'll update it to just link to the script in the repo

dennisnewel:
removed the script and updated the link to the scripts repo instead

PR #634, opened 2017-02-09
First draft Snap migration doc

ethangj:


ethangj:
Table does not work.

mlocher:
@ethangj We'd need to write HTML code for the tables (for now), see https://github.com/codeship/documentation/blob/master/_pro/getting-started/handling-secrets.md for an example.

PR #635, opened 2017-02-10
Update packages list

LarsFronius:
Like `dpkg -l | grep '^ii' | awk '{print $2 "\t" $3}' | column -t`


mlocher:
@LarsFronius could you create a new file with the AMI ID as the filename (in the same directory) and update the file at https://github.com/codeship/documentation/blob/master/_basic/getting-started/packages.md to point at the new list?

(Reason being that we sometimes need to revert to an older image and if we do need to revert we only have to change a single setting on the docs.)

LarsFronius:
Unfortunately I have no idea what the old AMI ID has been...
Do you know if the documentation repository is able to pull in information from S3? It should be simple to continuously keep a file updated with the latest package list from the checkbot master build.

mlocher:
I talked to @LarsFronius on Slack and we are going to work on automating this process. First task is to generate the list each time we build a new build VM AMI and upload it to S3.

PR #645, opened 2017-02-24
Service timing instructions added to Services page.

ethangj:


_pro/getting-started/services.md
```diff
@@ -208,7 +208,46 @@ All linking to the host is not allowed. This means the following directives are
   * `ports`
   * `stdin_open`
 
-Labels as they relate to images are supported by Codeship and should be [declared in the Dockerfile](https://docs.docker.com/engine/reference/builder/#/label) using the `LABEL` instruction. `labels` as a key in the services file (to label the running container) is not supported. 
+Labels as they relate to images are supported by Codeship and should be [declared in the Dockerfile](https://docs.docker.com/engine/reference/builder/#/label) using the `LABEL` instruction. `labels` as a key in the services file (to label the running container) is not supported.
+
+## Timing And Waiting
+
+One common issue Docker users encounter is around the `links` directive, outlined above and used to orchestrate dependent containers. A container orchestrated via `links` does not force the primary container to wait until uptime to proceed.
```
mlocher: until _it is ready_ maybe instead of until _uptime_?

_pro/getting-started/services.md
```diff
@@ -208,7 +208,46 @@ All linking to the host is not allowed. This means the following directives are
...
+
+There are two common solutions to this problem:
+
+- Adding a short `sleep` command before your tests, giving the dependent service slightly more time to start up before proceeding.
+
+- Writing a simple health poll script to check for service uptime before proceeding.
+
+An example of a health polll script may look something like this:
+
+```bash
```
mlocher: We can also link to https://github.com/codeship/scripts/blob/master/utilities/check_port.sh for a generic script to check for an open port (which is not always the correct solution, because ports can be open without the service being ready)

_pro/getting-started/services.md
```diff
@@ -208,7 +208,46 @@ All linking to the host is not allowed. This means the following directives are
...
+#!/usr/bin/env bash
```
mlocher: I'd like to take a look if that script requires `bash` or works with `/bin/sh` as well. `bash` is not available in all images.

_pro/getting-started/services.md
```diff
@@ -208,7 +208,46 @@ All linking to the host is not allowed. This means the following directives are
...
+
+function test_postgresql {
+  pg_isready -h "${POSTGRESQL_HOST}" -U "${POSTGRESQL_USER}"
```
mlocher: We should note that those tools (`pg_isready` in this test and `redics-cli` for the next one) need to be available in the image running the tests.

ethangj:
@mlocher up for re-review

PR #652, opened 2017-03-07
updating volume documentation to mention new limitations

dennisnewel:


_pro/getting-started/docker-volumes.md
```diff
@@ -14,6 +14,11 @@ redirect_from:
 {:toc}
 
 <div class="info-block">
```
mlocher: Should we think about a "warning-block" that isn't blue? Something similar to how Bootstrap (among others) handle this in their Documentation (e.g. http://getbootstrap.com/components/#progress the red and blue boxes)

ethangj: +1

dennisnewel: i was trying to make that work but couldn't find the right class. I'll try again to see if some of the other bootstrap ones will apply

mlocher: Bootstrap doesn't offer those elements (and afaik we are not using Bootstrap). We'd need to talk to Roman and see if he can add it (or provide us the instructions so we can add it)

dennisnewel: found the right class `alert-block` so now it's red :)

_pro/getting-started/services.md
```diff
@@ -121,6 +121,8 @@ data:
     - ./tmp/data:/data
 ```
```
LarsFronius: I believe we should state here, that ./tmp/data has to exist as a folder in the repository.

mlocher: @LarsFronius, there's an ongoing discussion on the announcement Google Doc about this, could you take a look?

Basically the gist is that many people are reluctant to add directories to their repositories they don't actually need. `tmp` being a good example that's usually `.gitignored`.

LarsFronius: Let's call it ./mnt/data then? :) 
On your concerns regarding host volumes to share data between steps - this is not an issue and will continue to work with relative paths like the one we have here.

mlocher: The main takeaway (or question) was whether we want to make such a directory available to the user or not. Basically do we make sure that a `tmp` directory (or `mnt/tmp`) is available on all builds even if the user doesn't include it in their repository. Or not. 

My reservation is that if we force users to add directories to their repositories that they only need for CI we will get pushback from people who want to keep their repositories clean and free of clutter.

dennisnewel: I think we should add a folder that people can use, but i don't think it's a blocker for the announcement to go out?
maybe call it `./codeshiptmp` to avoid any clashes with directories in the repo?

mlocher: I think it's fine to check for the `tmp` folder and just create it if it's not already there.

dennisnewel: ok, i'm going to add a new story to pivotal to auto-create a ./tmp folder if one doesn't exist after the clone has happened. I don't want it to stop the messaging to go out though, so it'll be something we'll add as soon as possible

PR #671, opened 2017-03-21
Move search to be always visible 

ethangj:


ethangj:
@mlocher released at staging/move-search for review

PR #674, opened 2017-03-22
correcting use of dockerfile_path key

dennisnewel:
we say that use of `dockerfile_path` is deprecated, but almost all examples used it 😄 
this PR should correct the examples


dennisnewel:
was caught by a customer 😄 

I can't merge, so i'll leave it up to one of you

PR #690, opened 2017-03-30
Standardizing all Basic language and framework articles

ethangj:


_basic/languages-frameworks/dart.md
```diff
@@ -20,32 +22,45 @@ You can use these commands from the Dart SDK:
...
-## Installing dependencies
+### Installing New Versions
 
-You can install any dependencies you defined in your pubspec.yaml by running:
+If you want to use a newer Dart version than what is pre-installed on the build  machines, you can use the following script:
+
+```shell
+wget https://storage.googleapis.com/dart-archive/channels/stable/release/latest/sdk/dartsdk-linux-x64-release.zip
+unzip -o dartsdk-linux-x64-release.zip -d ~
+```
```
mlocher: @joesiewert we should create a script for this. What do you think?

joesiewert: We've already got it actually:  https://github.com/codeship/scripts/blob/master/languages/dart.sh

Just need to update this doc to point to it I think.

PR #697, opened 2017-04-05
removing reference to old jet-steps files

dennisnewel:
These files are not really used anymore and cause confusion


dennisnewel:
took Marko off as reviewer; probably just need one 😄

PR #698, opened 2017-04-06
Fix links and update Node repo

joesiewert:
URLs should show as links now and changing the Node repo is based on customer feedback.


freewil:
how often is the documentation pushed to production? I still don't see these updates on the live site.

joesiewert:
Should be up now:  https://documentation.codeship.com/basic/builds-and-configuration/parallelci/#parallel-modules

PR #705, opened 2017-04-12
Sidebar updates

kellyjandrews:
This pull request modifies the sidebar navigation in several ways.

1.  **Side bar categories are now collapsable.** 

2. **Side bar categories are context aware**

3. **Sidebar categories are now controlled in the [`_config.yml` file](https://github.com/codeship/documentation/compare/master...sidebar-updates?expand=1#diff-aeb42283af8ef8e9da40ededd3ae2ab2R90)** 
4. **Categories have moved to [`_config.yml` defaults ](https://github.com/codeship/documentation/compare/master...sidebar-updates?expand=1#diff-aeb42283af8ef8e9da40ededd3ae2ab2R25)**
5. **Pro/Basic sidebar includes General categories**


kellyjandrews:
Ah yeah - I forgot I saw the small width issue. I can go back and clean it up. 

Not sure I would be able to split things up into 2 PRs here without manual effort.  I can try to make the PRs smaller in the future.

_config.yml
```diff
@@ -86,40 +165,40 @@ kramdown:
     line_numbers: false
 
 # Prose.io
```
mlocher: Need not happen in this PR, but I think we can throw out Prose for now. If we want to add it again, we'd need to start from scratch anyways.

_config.yml
```diff
@@ -22,6 +22,104 @@ defaults:
       weight: 50
       comments: true
       image: /assets/img/og_documentation.jpg
+  - scope:
```
mlocher: I'm personally not a fan of having the category defined outside of the YAML frontmatter. I know that with the directory structure we have right now it's easy to do this, but I prefer to have it assigned explicitly on the post itself.

ethangj: If the folder structure is always available, I'm not sure it makes a big difference, but if we could end up in a situation where we want to categorize separate from the folder structure then we should probably keep front matter in place.

ethangj:
@kellyjandrews @mlocher 

I'd go ahead and merge this. I created new cards assigned to Kelly to create category overviews and reintroduce frontmatter, and put a card on to investigate menus for a longer-term solution.

PR #708, opened 2017-04-13
Pro getting started consolidation

ethangj:
Initial consolidation of articles, refactor of articles next in line.


ethangj:
@mlocher local linter is not catching these path errors for me, can you verify?

to clarify, tons of path errors exist, but linter on my local machine is not catching them. too inefficient to use codeship to solve them, need to figure out why linter isn't working locally.

mlocher:
I get the same errors when I run the following commands

```
docker build --tag codeship/documentation .
docker run -it -p 4000:4000 -v $(pwd):/docs codeship/documentation
```

I'll take a look and see if I can resolve the merge conflicts (which is a good reason to keep each PR focused on just on set of changes) and fix the issues with the `jekyll build` as well.

mlocher:
I force pushed, to resolve the merge conflicts. 

I removed the changes to the `Gemfile` and `Gemfile.lock` (and replaced them with the changes from #707) and removed the `.ruby-version` file (as RVM will pick up the ruby version from the `Gemfile` (see https://rvm.io/workflow/projects#supported-files)

mlocher:
I also do get the same error (it will fail on the first one) when I run the commands locally

```
# Switch to Ruby 2.4.1, install it and create a 'documentation' gemset
rvm use 2.4.1@documentation --install --create
gem install bundler
bundle install

jekyll build
```

ethangj:
@mlocher should be resolved.

PR #709, opened 2017-04-13
Modifying skipping builds info.

ethangj:
~Removing misc basic article since info is contained in Getting Started guide. Adjusting Pro article with more info for now, will move it as well later.~

Better idea, moving skipping builds article from both basic and pro and moving to general collection.


ethangj:
@mlocher local linter is not catching these path errors for me, can you verify?

PR #715, opened 2017-04-18
updated link to API documentation

dennisnewel:
now pointing to custom domain


mlocher:
@ethangj The link in the header still uses the old URL. Imo it would make sense to extract the URL to the API Docs into a variable (in the `_config.yml`) and only reference it via the variable, whenever we need it.

PR #718, opened 2017-04-19
Further image articles consolidation

ethangj:


ethangj:
@mlocher Figured we should delete it because, since it was first added we've never updated it with any new images. I like the idea but don't think we should maintain an article around a single contribution from months ago.

PR #740, opened 2017-04-26
Fixing config logic that was overwriting redirects and leaving a lot …

ethangj:
Fixing config logic that was overwriting redirects and leaving a lot of 404s


mlocher:
Those redirects also redirect any subpages of `/classic/*` How sure are we that we aren't missing any of those?

ethangj:
@mlocher it would be redirecting almost all of them to 404s, though. can use 404 report to make sure nothing is missed.

PR #741, opened 2017-04-26
Buncha redirect fixes

ethangj:


ethangj:
Need to end the 404s for now, can always recrawl later if we reimplement category pages.

PR #743, opened 2017-04-27
Update paths to release notes in build.sh

GeorgeMac:
I believe this is where we want to shuttle the release notes?


GeorgeMac:
Nice I hope this does work!

PR #746, opened 2017-04-27
Screenshot swap and removing TOC from release log

ethangj:


_basic/builds-and-configuration/packages.md
```diff
@@ -25,7 +25,7 @@ In a best effort to list everything installed on  our hosted Codeship Basic envi
 
 If you want to check for a certain package or see what the default version is for yourself, you can use the [SSH Debug session]({{ site.baseurl }}{% link _basic/builds-and-configuration/ssh-access.md %}) within your project and connect directly to our build machine.
 
-![Open SSH Debug Session]({{ site.baseurl }}/images/basic/open_ssh.png)
+![Reset Dependency Cache]({{ site.baseurl }}/images/basic-guide/reset-dependency-cache.png)
```
mlocher: Can we either take a separate screenshot with the _Debug Build_ option highlighted or rename that screenshot.

I was stumped when I read that change and was wondering why you would do that. I had to check the current live site and the proposed new screenshot to make sense of it.

PR #748, opened 2017-04-28
Reinstating tag views, but broken build

ethangj:
@mlocher this reinstates tag views (which works!) but the build fails with url conflict errors related to tags. would be great if you could take a look at how to resolve these conflicts as i'm stuck (no rush)

### DIFF
--- a/_layouts/tags.html
+++ b/_layouts/tags.html
@@ -3,8 +3,6 @@ layout: default
 title: Documentation
 bodyclass: is-blue
 ---
-{% assign collections = site.collections %}
-
 <div class="contentWrapper">
   <div class="grid">
     <div class="row">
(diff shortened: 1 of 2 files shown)


mlocher:
@ethangj Took a look and reimplemented the logic. The problem was that you created the same tag overview for each page that included a specific tag. So if 3 documents included the `aws` tag for example, you create the `tags/aws/index.html` page 3 times, causing those conflicts.

I changed the logic to iterate over all documents for all collections and assembling a hash of pages belonging to specific tags, e.g.

```
{
  'aws': [
    document1,
    document2,
    ...
  ],
  ...
}
```

This hash is then used to generate the individual tag pages. I've also added logic to pass the tags directly to those pages, cutting down on the logic in the tag pages itself.

We might want to look at sorting those pages again (or adding the additional logic that was stripped on the tag overview pages).

mlocher:
I tested my changes locally, but we should definitely run a staging build for this :)

ethangj:
Awesome, thanks for fixing that. Looks good on https://documentation.codeship.com/staging/tags-release/tags/deployment/

PR #764, opened 2017-05-09
Setup test commands

ethangj:
This is just adding existing commands

Card created to review and update as needed and add missing commands

Probably need to rebase once Python changes merged

### DIFF
--- a/_basic/languages-frameworks/dart.md
+++ b/_basic/languages-frameworks/dart.md
@@ -31,6 +31,16 @@ wget https://storage.googleapis.com/dart-archive/channels/stable/release/latest/
 unzip -o dartsdk-linux-x64-release.zip -d ~
 ```
 
+### Default Setup Commands
+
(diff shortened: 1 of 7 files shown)


mlocher:
I just started working on improving the current command templates (see codeship/mothership#4788), linking each of the templates to the respective documentation page.

If we want to have setup and test commands on the documentation pages I would wait for the above PR to be finished. And then implement automation to pull the data from the application repository itself.

ethangj:
@mlocher Awesome. Is that automation something you can handle, just using the same format as the version in this PR?

mlocher:
@ethangj we'd need to take a look. My current estimate is that we would need to write a script similar to the `jet` CLI changelog script. 

We'd also need to think about which command templates to show, especially in regards to frameworks. (E.g. on the Ruby article, show we only the generic Ruby command template, or the templates for Rails and Sinatra as well).

PR #782, opened 2017-05-22
Add the Zendesk widget to documentation pages

mlocher:
Widget and enabled features are configurable via the _config.yml file.

At the moment the widget is enabled, but Zendesk Chat is not.


mlocher:
Thanks for taking the time to trigger a staging build. I'll create a follow up card to see if we can automatically sign in users on the widget if they are signed in on Codeship.

We can then also discuss enabling Chat on the documentation.

PR #790, opened 2017-05-30
Update packages for the latest AMI

mlocher:
Update the package and languages list for the latest AMI. Also make the script to generate those lists a bit easier to use, switch to the [unofficial bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/#issues-and-solutions) (for most of the script) and print a sample `scp` command to download the generated archive.


ethangj:
@mlocher still lgtm but not sure if you think its worth getting a second opinion from someone a bit more savvy on the script itself.

PR #793, opened 2017-05-31
No analytics

ethangj:


.ruby-version
```diff
@@ -0,0 +1 @@
+2.4.1
```
mlocher: Do we need that file? It makes updating to a new Ruby version more complicated since it's now two files (the `Dockerfile` and the `.ruby-version` file) that need to be updated.

I personally would prefer removing it again and instead making the Docker based workflow easier to follow on the development machines as well.

ethangj:
@mlocher removed stupid ruby version file, should be g2g now

PR #796, opened 2017-06-04
Percy fix

ethangj:


_pro/continuous-integration/percy-docker.md
```diff
@@ -28,7 +28,7 @@ To use Percy with static sites inside Docker images on Codeship Pro, you will ne
 RUN gem install percy-cli
```
mlocher: Should the gem name up here be updated to `percy-capybara` as well?

ethangj: @mlocher i don't know actually, these corrections came from Percy and they only mentioned it in relation to this one instance, not the gem install.

PR #811, opened 2017-06-07
Adding signup link to end content block

ethangj:
Per marketing request


mlocher:
Imo, that link doesn't really fit there: *Need More Help* --> _Sign Up Or Login_. And we already have a large _Sign Up_ link in the top navigation.



(The screenshot is a fake, so there might be slight variations in the wording, but afaik the position is correct).

PR #815, opened 2017-06-09
 Css Overhaul

ethangj:
Staged at http://documentation.codeship.com/staging/cssredo

Feedback on these changes would be great. Standardizing some colors and type and sizing with new design, other misc little tweaks. 

Will keep this PR open for a bit as I add tweaks to it


ethangj:
@mlocher this is now updated with fixes from roman, mobile fixes and i think pretty close to as good as i can get on fonts and spacing. 

can we come up with a final punch list to get this shipped, even if we want to do a round two later?

PR #818, opened 2017-06-13
Keyboard shortcuts

ethangj:


_general/account/new-user-signup.md
```diff
@@ -134,3 +134,10 @@ git config --get user.email
 ```
 
 Note that different projects can have different email addresses configured and that your VCS can have other email addresses configured for the actions you take via their interfaces.
+
+To make it easier to navigate through your Codeship projects, we've provided several keyboard shortcuts for quickly jumping through the interface:
+
+- `gp` to view all of your projects
+- Escape key, inside a project, to go back to viewing all projects
```
alextacho: Those keyboard shortcuts are only available on Projects Overview page:
- `Escape` key on the /projects page will go back to previous page
- `Arrow keys` on /projects page let you select a project
- `Enter` confirms project selection and will go to that project

ethangj:
@alextacho updated

alextacho:
looks good one addition is that "gp" is actually available on all pages. might be covered with your content though.

PR #821, opened 2017-06-14
Codeclimate does parallel now

ethangj:


_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
 
 ### Application Configuration
 
-Once your API key is loaded, you will want to configure Code Climate to run inside your application, during your test, as you would normally without any additional modification.
+Once your Code Climate project ID is loaded via your environment variables, you will need to add special Code Climate commands before and after your [test commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}):
+
+Before your tests:
+
+```bash
+cc-test-reporter before-build
```
mlocher: Installation instructions for the new test reporter are missing

```
curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > "${HOME}/bin/cc-test-reporter"
chmod +x "${HOME}/bin/cc-test-reporter"
```

we also might want to add a [codeship/scripts](https://github.com/codeship/scripts) script for this.

_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
+```
+
+After your tests:
+
+```bash
+cc-test-reporter after-build
```
mlocher: I'm not sure if we need to run this after each pipeline, or only once per build. If we only need to run it once per build, where do we run it to make sure all pipelines are finished? A deployment would be a good step, but we can't run a deployment on each build :(

_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
+### Parallel Test Coverage
+
+Code Climate supports parallel test reports, as a new beta feature, by uploading the partial result to an external CDN. In addition to the pre-test and post-test commands up, to use Code Climate with parallel reporting you will need to add another command at the end of your [test commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}), in each [parallel pipeline]({{ site.baseurl }}{% link _basic/builds-and-configuration/parallelci.md %}) that you run tests in - as well as a new command at the end of your build.
+
+Here are [Code Climate's example](https://github.com/codeclimate/test-reporter#low-level-usage) scripts for doing so.
+
+At the end of each [parallel pipeline]({{ site.baseurl }}{% link _basic/builds-and-configuration/parallelci.md %}):
+
+```bash
+./cc-test-reporter format-coverage --output "coverage/codeclimate.$N.json"
```
mlocher: `$N` is not defined. Afaik we don't make the pipeline number available as an environment variable on Basic yet, but this would be helpful. If we don't have the pipeline number (or name), we need to find another unique value instead of `$N`

_basic/continuous-integration/code-climate.md
mlocher: Also, we should probably make use of the `--exit-code` switch

```
cc-test-reporter after-build --exit-code $?
```

_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
+
+After your tests:
+
+```bash
+cc-test-reporter after-build
+```
+
+### Parallel Test Coverage
+
+Code Climate supports parallel test reports, as a new beta feature, by uploading the partial result to an external CDN. In addition to the pre-test and post-test commands up, to use Code Climate with parallel reporting you will need to add another command at the end of your [test commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}), in each [parallel pipeline]({{ site.baseurl }}{% link _basic/builds-and-configuration/parallelci.md %}) that you run tests in - as well as a new command at the end of your build.
```
mlocher: S3 is not a CDN, it's a storage service

_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
+
+Here are [Code Climate's example](https://github.com/codeclimate/test-reporter#low-level-usage) scripts for doing so.
+
+At the end of each [parallel pipeline]({{ site.baseurl }}{% link _basic/builds-and-configuration/parallelci.md %}):
+
+```bash
+./cc-test-reporter format-coverage --output "coverage/codeclimate.$N.json"
+aws s3 sync coverage/ "s3://my-bucket/coverage/$SHA"
```
mlocher: I would replace `$SHA` with a unique value we have available, either the commit hash or the build ID (possibly extended by a timestamp to account for re-run builds)

_basic/continuous-integration/code-climate.md
```diff
@@ -17,20 +17,61 @@ redirect_from:
...
+```
+
+Note that you will need to modify the S3 path (or provide an alternative CDN), as well as the `$SHA` and the `$N` value.
+
+At the end of your build itself, as a new test command placed after your normal tests:
+
+```bash
+cc-test-reporter sum-coverage --output - --parts $PARTS coverage/codeclimate.*.json | \
```
mlocher: Same as above, how do we make sure this is run after all pipelines are finished?

mlocher: we can also replace `$PARTS` with the number of parallel pipelines on Basic, which is available as an environment variable as far as I know.

_pro/continuous-integration/codeclimate-docker.md
```diff
@@ -13,18 +13,88 @@ menus:
...
+Starting with Code Climate and Codeship is easy. [Their documentation](http://docs.CodeClimate.com/article/219-setting-up-test-coverage) do a great job of guiding you, but the first step is to add your `CC_TEST_REPORTER_ID` to the [encrypted environment variables]({{ site.baseurl }}{% link _pro/builds-and-configuration/environment-variables.md %}) that you define in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
 
 ### Application Configuration
 
-Once your API key is loaded, you will want to configure Code Climate to run inside your application, during your test, as you would normally without any additional modification.
+Once your Code Climate project ID is loaded via your environment variables, you will need to add a couple additional commands to your pipeline via your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}).
+
+Before your test commands:
```
mlocher: We're missing installation commands (or at least pointers) for how to get the `cc-test-reporter` binary into the test container.

mlocher: Also, just running the below two steps will result in empty coverage, as each container does not persist state from previous steps.

We need to look into how to share the required state between those steps or switch to a script based solution run in a single step instead.

mlocher:
I would consider removing the Pro article for now, and merging Basic to get it out and afterwards tackle Pro in another PR.

_basic/continuous-integration/code-climate.md
```diff
@@ -34,42 +34,49 @@ You can do this by navigating to _Project Settings_ and then clicking on the _En
...
 At the end of each [parallel pipeline]({{ site.baseurl }}{% link _basic/builds-and-configuration/parallelci.md %}):
 
 ```bash
 ./cc-test-reporter format-coverage --output "coverage/codeclimate.$N.json"
-aws s3 sync coverage/ "s3://my-bucket/coverage/$SHA"
+aws s3 sync coverage/ "s3://my-bucket/coverage/$CI_COMMIT_ID"
 ```
 
-Note that you will need to modify the S3 path (or provide an alternative CDN), as well as the `$SHA` and the `$N` value.
+Note that you will need to modify the S3 path (or provide an alternative storage path), as well as set the `$N` value value by manually declaring the number of pipelines.
```
mlocher: `$N` is the current pipeline, not the number of used pipelines.

_basic/continuous-integration/code-climate.md
```diff
@@ -34,42 +34,49 @@ You can do this by navigating to _Project Settings_ and then clicking on the _En
...
 
-At the end of your build itself, as a new test command placed after your normal tests:
+At the end of your build itself, as a command run via a the [custom-script deployment option]({{ site.baseurl }}{% link _basic/continuous-deployment/deployment-with-custom-scripts.md %}):
```
mlocher: But how do we configure the deployment to run on each build? Wildcard deployments don't support this at the moment.

_pro/continuous-integration/codeclimate-docker.md
```diff
@@ -28,7 +28,14 @@ Starting with Code Climate and Codeship is easy. [Their documentation](http://do
 
 ### Application Configuration
 
-Once your Code Climate project ID is loaded via your environment variables, you will need to add a couple additional commands to your pipeline via your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}).
+Once your Code Climate project ID is loaded via your environment variables, you will need to install Code Climate into one of your services via your Dockerfile by using the following command:
+
+```bash
+curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > "${HOME}/bin/cc-test-reporter"
```
mlocher: we shouldn't use `$HOME/bin` because that is Codeship Basic specific. On Docker we can use `/usr/local/bin/` instead.

_pro/continuous-integration/codeclimate-docker.md
```diff
@@ -38,47 +45,41 @@ Before your test commands:
...
+      command: tests1.sh
     - service: YOURSERVICE
-      command: test_commands
-    - service: demo
-      command: codeclimate-post.sh  
+      command: tests2.sh
 ```
 
-Note that we're using serial step groups being run in parallel, s that we can run our script after each parallel test thread completes. Inside the new `codeclimate-post.sh` file, you will have:
+Note that we're using script filesto run our tests, so that we can execute the tests and export the coverage report as one command. This is because each step uses a new container, so the coverage report will not persist if the commands are separated. Inside the new `tests.sh` files, you will have:
```
mlocher: script _files to_ (missing whitespace)

_pro/continuous-integration/codeclimate-docker.md
```diff
@@ -38,47 +45,41 @@ Before your test commands:
...
 
 ```bash
+# your test commands go here
+
 ./cc-test-reporter format-coverage --output "coverage/codeclimate.$N.json"
```
mlocher: that should probably be just `cc-test-reporter`, depending on where the binary is installed to in the first place.

_pro/continuous-integration/codeclimate-docker.md
```diff
@@ -38,47 +45,41 @@ Before your test commands:
...
-aws s3 sync coverage/ "s3://my-bucket/coverage/$SHA"
+aws s3 sync coverage/ "s3://my-bucket/coverage/$CI_COMMIT_ID"
 ```
 
-Note that you will need to modify the S3 path (or provide an alternative CDN), as well as the `$SHA` and the `$N` value.
+Note that you will need to modify the S3 path (or provide an alternative storage path), as well as set the `$N` value value by manually declaring the number of pipelines.
```
mlocher: `$N` is the current parallel step ID, not the max number of parallel steps

PR #830, opened 2017-06-19
OpsGenie draft

ethangj:
@mlocher Do you know if we have, or if you could quickly write, a very simple script that just shows something like a simple if with a real deployment command (like s3 or heroku)? I think it's optional in the scope of this, and I don't see one sitting around, was just curious if we had anything prepared already to bump this up a bit in depth.

### DIFF
--- a/_general/integrations/opsgenie.md
+++ b/_general/integrations/opsgenie.md
@@ -1,0 +1,70 @@
+---
+title: Using OpsGenie To Track Deployments With Codeship Pro
+shortTitle: Tracking Deployments With OpsGenie
+menus:
+  general/integrations:
+    title: Using OpsGenie
+    weight: 6
+tags:
+  - opsgenie
+  - deployment
+  - logging
+  - analytics
+  - reports
+  - reporting
+  - integrations
+---
+
+* include a table of contents
+{:toc}
+
+## About OpsGenie
+
+[OpsGenie](https://www.opsgenie.com) lets send alerts and notifications about critical events to your on-call or development team.
+
+[Their documentation](https://www.opsgenie.com/docs) does a great job of providing more information, in addition to the setup instructions below.
+
+## Codeship Pro
+
+### Setting Your API Key
+
+You will need to add your OpsGenie API key to your [encrypted environment variables]({{ site.baseurl }}{% link _pro/builds-and-configuration/environment-variables.md %}) that you define in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
+
+### Sending Alerts
+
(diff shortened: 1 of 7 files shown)


mlocher:
Assuming you want to create a script to create an OpsGenie alert, they have a basic `curl` command available at https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest

We could either add that (in a modified way) to the documentation, or if we want to, add it to codeship/scripts as well.

ethangj:
@mlocher what about the conditional, showing only run the curl if something else fails? do we have an example on hand for that? probably not strictly necessary.

mlocher:
@ethangj closest we get is https://github.com/codeship/scripts/blob/master/utilities/ensure_called.sh

ethangj:
@mlocher kk, will just go with the opsgenie example then, can you go ahead and review PR

_general/integrations/opsgenie.md
```diff
@@ -0,0 +1,96 @@
...
+  - reporting
+  - integrations
+---
+
+* include a table of contents
+{:toc}
+
+## About OpsGenie
+
+[OpsGenie](https://www.opsgenie.com) lets send alerts and notifications about critical events to your on-call or development team.
```
mlocher: lets _you_ send ...

_general/integrations/opsgenie.md
```diff
@@ -0,0 +1,96 @@
...
+
+[Their documentation](https://www.opsgenie.com/docs) does a great job of providing more information, in addition to the setup instructions below.
+
+## Codeship Pro
+
+### Setting Your API Key
+
+You will need to add your OpsGenie API key to your [encrypted environment variables]({{ site.baseurl }}{% link _pro/builds-and-configuration/environment-variables.md %}) that you define in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
```
mlocher: that sentence sounds like it's missing something? The encrypted environment variables are not defined in the services file?

_general/integrations/opsgenie.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Sending Alerts
+
+Next, you will need to add [API calls to OpsGenie](https://www.opsgenie.com/docs/rest-api/alert-api), in your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}), most likely to alert of a successful deployment or combined with your deployment commands in a script to alert of a failed deployment.
```
mlocher: I would simplify that sentence and split it in two, one for each of the possible ways to integrate this.

PR #832, opened 2017-06-20
LFS upgrade

ethangj:


_general/integrations/git-lfs.md
```diff
@@ -2,12 +2,15 @@
 title: Support For Git LFS Codeship
 shortTitle: Git LFS
 menus:
-  general/about:
+  general/integrations:
```
mlocher: Not sure if we want to move that to the integrations. I'd rather we remove the need for the article completely and get both Basic and Pro to support LFS natively.

Until then, leave it in General/About? What do you think?

ethangj:
@mlocher reverting category

PR #844, opened 2017-06-29
Security page update

ethangj:


_general/about/security.md
```diff
@@ -47,19 +47,23 @@ Our whole infrastructure is based on Amazon EC2 or services built on top of it.
...
++ Looker
 + Source code access
 
 As outlined in our Terms of Service we only access your source code for a build or support request. We do not have any way to access your repository outside of our build environment.
 
-## How Does Codeship Back Up Data?
+## Does Codeship Save My Code?
 
-Codeship never takes ownership of a customer's files, with the exception of opt-in image caching on Codeship Pro. However, all Codeship file storage and backup is held in an encrypted state off-site.
+Codeship never takes ownership of your code or files. All builds run on containers or machines that are shut down at the end of your build, with your cloned repository and generated assets never persisting between builds.
```
mlocher: > assets never _persisted_ between builds. 

(instead of _persisting_, at least I think it should be that way, maybe it just sounds strange to me because it's a way to put this, that I haven't heard so far)

_general/about/security.md
```diff
@@ -47,19 +47,23 @@ Our whole infrastructure is based on Amazon EC2 or services built on top of it.
...
+
+The one exception is with opt-in caching. On Codeship Basic we will save your dependencies and on Codeship Pro we will save your images - but only if you explicitly tell us to in your project configuration.
```
mlocher: that sentence can be misleading, there is no way to opt out of the dependency cache on Basic, but the dependency cache on Basic doesn't store any private information either.

PR #847, opened 2017-06-30
Update Basic frontend testing documentation

joesiewert:


ethangj:
@joesiewert what do you think about adding a Troubleshooting or Common problems section with the first few troubleshooting steps we recommend?

PR #851, opened 2017-07-04
Update registry info regarding Docker Hub

rheinwein:
- Since we use the Docker client, the Hub is the default registry. A user doesn't have to specify a registry URL.
- Declaring the Hub as the default registry for pushes and pulls makes sense -- we don't make users specify a registry URL when pulling from the Docker Hub, so why on push?
- Pushing to a locally running registry isn't super important, but could be useful for a dry run of the push. I moved it down so it's not the first thing that users read. I also updated that repo to use the V2 registry image.


rheinwein:
@ethangj I'm updating this to make a few changes to consolidate instructions for generating encrypted credentials. I noticed that in every example for pulling a private base image, we instruct the user to update their steps file with the encrypted dockercfg file. While this technically works, it probably makes more sense to instruct them to add the dockercfg file to the service instead, since then they only have to add it once. Having the dockercfg on a push step makes total sense, but it might be a little confusing (and unnecessarily repetitive) to ask for credentials on a per-step basis for a private base image. Any objections to updating the docs to reflect that?

ethangj:
@rheinwein does it actually work without it on the push step?

rheinwein:
@ethangj yeah -- so long as it's defined on the service

PR #853, opened 2017-07-04
Removing parallelci from docs

ethangj:
Pending marketing launch.


index.html
```diff
@@ -1,4 +1,4 @@
----
+basic//index.md---
```
mlocher: That line should probably be reverted to `---` again.

PR #863, opened 2017-07-12
Adding category to Segment event

ethangj:
Tested locally for js errors and verified data coming through segment as intended.


_includes/segment.html
```diff
@@ -1,6 +1,8 @@
 <script>
   !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="4.0.0";
   analytics.load("PQiyLlOaOXibLK1hhXxBWR1iIJ20FLXl");
-  analytics.page();
+  analytics.page(`{{ page.title }}`, {
+     category: `{{ page.collection }} documentation`
```
mlocher: According to https://segment.com/docs/sources/website/analytics.js/#page the `page` function supports a `category` as the first parameter:

> `analytics.page([category], [name], [properties], [options], [callback]);`

Why aren't we making use of that?

ethangj: @mlocher Didn't work, it assigned it to `name` instead when passed in the single value. Then I passing it two values, for category and name in order as it indicates, it just cat'd them together still as `name`. By setting it this way I was able to get the result we wanted.

mlocher: That is weird, could you open a ticket with them to clarify, could be a bug on their end.

PR #869, opened 2017-07-13
Integrations page

ethangj:
CSS changes still need to reflect with cache clear, but staging build here: https://documentation.codeship.com/staging/integrations/integrations/

Let's do code modifications here and structure modifications if any on the Trello card: https://trello.com/c/hntggGzk/283-add-integrations-grid-page


mlocher:
Looks like this for me, is this accurate?

ethangj:
@mlocher 

Difference between local and staging for me:

screen shot 2017-07-13 at 10 41 53 am

screen shot 2017-07-13 at 10 41 56 am

google1a85968316265362.html
```diff
@@ -1 +0,0 @@
-google-site-verification: google1a85968316265362.html
```
mlocher: Why did we remove that file?

ethangj: Afaik we only need that file in there once, when Google initially verifies? I didn't think Google required it to permanently live on the site.

mlocher: Not sure, but I know other sites which regularly check that file. Is there a negative impact if we keep it?

ethangj:
@mlocher Google file back. Staging/prod are rendering relative fonts slightly different sizes than locally, but got it how I want on staging now.

PR #875, opened 2017-07-13
All images smaller

ethangj:


ethangj:
@mlocher Used a couple image compression libraries to reduce filesizes. Not a huge difference, I think 25mb or so reduced out of 70 or 80mb total,  but starting on th speed optimization checklist.

mlocher:
Yeah, just saw the trello card. Would be great if we could get that into the build pipeline, img size in the repository isn't as big of an issue as on the deployed site, and if it's automated we don't need to do it manually and can't forget :)

The `Gulpfile` should already contain starting points for this.

PR #881, opened 2017-07-14
add line for `-prefix` flag in Code Climate docs

davehenton:


davehenton:
Hi @mlocher and @ethangj 

This is Dave with Code Climate. We've found that this `-prefix` flag is necessary for our Codeship users. What can I do to help this update get merged? Let me know! Thanks.

mlocher:
Hey @davehenton, 

sorry for simply closing this (and thanks for putting in the time to prepare this PR in the first place).

Is there a way to define the prefix in a more dynamic way, e.g. by using `$(pwd)` or something similar? Reason I ask is that the path you added to the docs is only valid for Codeship Basic projects (and even there requires customization). 

For Codeship Pro we can't tell which path the user should add, since they can add the source to an arbitrary path within their Docker image. 

But if we can use a shell variable we should be fine in most cases.

davehenton:
Hey @mlocher, 

Thanks for the clarification there -- I agree with your point. We've added a CTA to our docs over [here](https://docs.codeclimate.com/docs/codeship-ci-test-coverage-example) that better explains the use of `-prefix`. Additionally, we'd like to have your docs on this integration point to the docs that we're hosting, so as to not actively maintain them in 2 different places. I'll close out this PR, and open a new one with that link. Thanks!

PR #883, opened 2017-07-17
Rancher draft

ethangj:


_general/integrations/rancher.md
```diff
@@ -0,0 +1,95 @@
...
+---
+
+* include a table of contents
+{:toc}
+
+## About Racher
+
+Rancher is a container management platform that helps bridge the gap between container stacks and infrastructure platforms. [Their documentation](http://rancher.com/docs/rancher/v1.6/en/) does a great job of providing more information, in addition to the setup instructions below.
+
+**Note** that Rancher only integrates with [Codeship Pro](https://codeship.com/features/pro) and will not work with [Codeship Basic](https://codeship.com/features/basic).
```
mlocher: Can we do that as an info block?

_general/integrations/rancher.md
```diff
@@ -0,0 +1,95 @@
...
+
+If you do not have a familiarity with Codeship Pro, we recommend watching this [introductory webinar](https://resources.codeship.com/webinars/env-parity-docker-codeship-jet) before proceeding with your Rancher setup.
+
+## Using Rancher
+
+We will not cover Rancher-side setup for your application in this documentation article, but if you are looking for more information on using Rancher itself we have a comprehensive [post on our blog](https://blog.codeship.com/deploying-rancher-using-codeship-pro/)
+
+You can also learn more from [their documentation](http://rancher.com/docs/rancher/v1.6/en/).
```
mlocher: Is there a way we can link to the latest version of the docs, instead of 1.6? http://rancher.com/docs/ redirects afaik.

_general/integrations/rancher.md
```diff
@@ -0,0 +1,95 @@
...
+## Defining Your Service
+
+Because all the commands in your pipeline, via your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}), are executed inside the service you define and build build via your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}) - the first thing you will need to do is define a service that is capable of executing Rancher commands (and, specifically Rancher Compose commands.)
+
+You can use an [existing image with Rancher compose configured](https://hub.docker.com/r/bfosberry/rancher-compose/) or build your own with a Dockerfile that looks similar to the one below:
+
+```bash
+FROM debian:8.1
+
+ENV RANCHER_COMPOSE_VERSION v0.12.0
```
mlocher: Should that be a build argument instead of an environment variable?

mlocher: And should we maybe add that image to the `codeship-library` github account and our Docker Hub account?

_general/integrations/rancher.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Deploying With Rancher
+
+After creating your encrypted Rancher keys and defining a service to execute your Rancher Compose commands, you will now want to add those commands to your pipeline via your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}):
+
+
+```bash
+- name: rancher-deploy
+  service: rancher
+  command: "rancher-compose  --p $STACK_NAME --verbose up -d --force-upgrade --pull --confirm-upgrade $SERVICE_NAME"
```
mlocher: Need to be careful with anything that looks like shell variables, as they will not work that way. Maybe we should remove the `$` signs and prepend `YOUR_`?

_general/integrations/rancher.md
ethangj: @mlocher Brendan's blog post says env vars, not build args - based it on that. I worry about hosting the image ourself, too, as far as maintanence goes. We can swap his out with another if his breaks, but once we host one there's a certain expectation on us.

mlocher: Agreed, we might want to look into hosting one if we get requests for it.

PR #891, opened 2017-07-19
Public key clarification Gcloud

ethangj:
Amongst the other refactoring I forgot the actual original task.

### DIFF
--- a/_pro/continuous-deployment/google-cloud.md
+++ b/_pro/continuous-deployment/google-cloud.md
@@ -92,6 +92,12 @@ codeship_google authenticate
 
 Since this authentication does not persist between steps in your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}), you will need to run the provided authentication command at the beginning of each step that attempts to run commands via the Google Cloud deployment container.
 
+### Codeship Public Key
+
+Some Google Cloud services will require that you add your [Codeship public key]({{ site.baseurl }}{% link _general/projects/project-ssh-key.md %}) for authentication purposes.
+
+**Note** that Google may fail authentication if you do not add the Google Cloud user the key is for to the end of the key. For example, if the Google Cloud user is `deploy@Codeship`, you will want to add `deploy@Codeship` to the end of the SSH key itself, otherwise Google will not load the key for the user appropriately.
+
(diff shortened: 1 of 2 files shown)


_pro/continuous-deployment/google-cloud.md
```diff
@@ -92,6 +92,12 @@ codeship_google authenticate
 
 Since this authentication does not persist between steps in your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}), you will need to run the provided authentication command at the beginning of each step that attempts to run commands via the Google Cloud deployment container.
 
+### Codeship Public Key
+
+Some Google Cloud services will require that you add your [Codeship public key]({{ site.baseurl }}{% link _general/projects/project-ssh-key.md %}) for authentication purposes.
+
+**Note** that Google may fail authentication if you do not add the Google Cloud user the key is for to the end of the key. For example, if the Google Cloud user is `deploy@Codeship`, you will want to add `deploy@Codeship` to the end of the SSH key itself, otherwise Google will not load the key for the user appropriately.
```
mlocher: Should that be an info block, or do you think a standard paragraph is fine?

_pro/continuous-deployment/google-container.md
```diff
@@ -72,6 +72,12 @@ jet encrypt your_env_file your_env_file.encrypted
 
  This encrypted file will be committed to your repository and used in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
 
+### Codeship Public Key
+
+Some Google Cloud services will require that you add your [Codeship public key]({{ site.baseurl }}{% link _general/projects/project-ssh-key.md %}) for authentication purposes.
+
+**Note** that Google may fail authentication if you do not add the Google Cloud user the key is for to the end of the key. For example, if the Google Cloud user is `deploy@Codeship`, you will want to add `deploy@Codeship` to the end of the SSH key itself, otherwise Google will not load the key for the user appropriately.
```
mlocher: same question as above?

ethangj: @mlocher I'd leave as is, I worry that info block in this case would make it seem "extra" rather than the main point? Not sure if that makes sense.

PR #917, opened 2017-07-27
JS Audit

mlocher:
Re-evaluate included JS scripts to speed up page load times

* Remove CrazyEgg
* Remove Google Analytics


ethangj:
@mlocher Why are we removing GA?

mlocher:
@ethangj GA is handled via Segment, so we were including it twice previously. I'm talking with Manny about migrating some of the other JS to Segment as well, as is overall easier to manage.

PR #919, opened 2017-07-27
Introduce Shipyard

ryanjwilke:
This PR is aiming to integrate the shipyard gem in order to start sharing styles across the app.


ryanjwilke:
@ethangj Would love your eyes on this. I've got the base styles included for Shipyard now and I we should be able to integrate things little by little now with this implementation. (e.g. typography first, then buttons, etc...)

ryanjwilke:
FYI, I could use your help clicking through a bunch of pages and making sure nothing critical is broken.

ethangj:
@ryanjwilke Nothing looks changed, all looks like it should.  Would like @mlocher to give this a run through as well, though.

mlocher:
This is adding quite a lot of dependencies that we don't make any use of with the site (like `actionpack`, `actionview`, `activesupport`, `rack`, `railties`, ...

Is there any way we can get those dropped again? E.g. does `shipyard-framework`, need to depend on `slim-rails` and `sassc-rails`, which pull in all of the other dependencies? Or could we potentially publish a stripped down version without those dependencies?

ryanjwilke:
@mlocher Yep. Good point, I'll figure out how to trim those down before this goes out.

ryanjwilke:
@mlocher This latest update should resolve your concerns.

mlocher:
@ryanjwilke awesome, thanks for making those changes. Imo this is ready to get merged now 🙇  :shipit:

PR #926, opened 2017-07-31
Link fix

ethangj:


_sass/docs/_page.scss
```diff
@@ -361,7 +361,7 @@ div#integrations {
   }
 
   a {
-    color: $_cDarkblue;
+    color: #fff !important;
```
mlocher: do we need the color in here (and do we want to change it from dark blue to white).

Same with the `!important` is this necessary to get this working?

E.g., this is going to switch the colors for links in the info, warning and feedback blogs

ethangj: @mlocher I do think the color should be white, it's very hard to read the dark link color against the gray background of end block and blue background of info block right now.

mlocher: Ok, then let's fix that in the overrides of those blocks and not via the default `a` tag. See e.g. https://github.com/codeship/documentation/blob/master/_sass/docs/_page.scss#L382-L384 for the feedback block (and above for the others).

If we make those links white, we need to highlight them in a different way though, which is the main reason they are not white right now.

ethangj: @mlocher It's not the default a tag, it's the a tag specifically for the info and end content blocks. This is how it was set prior to the design updates you made.

Just made the links bold, though, so that they'd be more visible against the paragraph font.

mlocher: Ok, in that case we shouldn't need the `!important`, `#fff` should work on it's own. And if it doesn't, we'd need to look into why it doesn't work.

Also, this will then not update the `feedback-block` and `warning-block` colors for links, which will still be dark grey and dark red respectively.

ethangj: @mlocher Updated with no important Working on a different PR for end content block.

PR #941, opened 2017-08-03
Update PHP versions

joesiewert:


mlocher:
Should we remove the patch levels from the mentioned versions? We have the exact version numbers available on the packages document, which get's updated regularly?

joesiewert:
Agree, let's chat as a team about a better way of handling the different versions of things across the site.

mlocher:
Since we have version files available for all languages we support via a version manager, we could take a look at using those to automatically provide a list of minor versions we support.

PR #959, opened 2017-08-21
Gemnasium

ethangj:


_general/integrations/gemnasium.md
```diff
@@ -0,0 +1,99 @@
...
+---
+
+* include a table of contents
+{:toc}
+
+## About Gemnasium
+
+[Gemnasium](https://gemnasium.com) is a service for analyzing and monitoring the security of your application dependencies.
+
+[Their documentation](https://github.com/gemnasium/toolbelt) does a great job of providing more information, in addition to the setup instructions below.
```
mlocher: Would prefer a more descriptive link text (but this is not a blocker)

_general/integrations/gemnasium.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Installing The CLI
+
+To use Gemnasium in your CI/CD process, you'll need to add the Gemnasium CLI to a service in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
+
+To add the Gemnasium CLI, you will need to add the following command to the Dockerfile for the service you want to run Gemnasium on:
+
+
+```bash
+sudo apt-get install gemnasium-toolbelt
```
mlocher: That command is only valid for Debian based distributions, and assumes that the package is available in the default repositories. Should we instead link to their installation instructions?

ethangj:
@mlocher Updated

PR #960, opened 2017-08-22
Hakiri

ethangj:


_general/integrations/hakiri.md
```diff
@@ -0,0 +1,119 @@
...
+To use Hakiri in your CI/CD process, you'll need to add the [Hakiri Toolbelt](https://github.com/hakirisec/hakiri_toolbelt) to a service in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
+
+To install the Hakiri Toolbelt, you will need to add the following command to the Dockerfile for the service you want to run Hakiri on:
+
+
+```bash
+gem install hakiri
+```
+
+**Note** that this requires the Dockerfile to also have Ruby, Rails and Ruby Gems installed.
```
mlocher: Do we require Rails, or just Ruby and the `gem` binary?

PR #966, opened 2017-08-24
Updates to Ruby and Node.js dependencies

mlocher:


greysteil:
Would you guys be up for automating these kind of dependency updates? I built [Dependabot](https://dependabot.com) to do that while I was at my last company - might save you a bit of time?

mlocher:
Thanks for pointing out the service, @greysteil. We're going to take a look and might integrate it in the future. I think for now the current workflow is still ok, especially for the documentation project :)

PR #968, opened 2017-08-24
Upgrade shipyard to 0.4.4.

ryanjwilke:
Upgrades to the latest version of Shipyard which also makes use of some of the lovely Rails helpers available for `buttons` and `alerts`.


mlocher:
Let's merge #966 first (which upgrades other dependencies) and then merge this PR.

mlocher:
Let's merge #966 first (which upgrades other dependencies) and then merge this PR.

It would still be great if we wouldn't need the `ActiveSupport` and `ActionView` dependencies, as they pull in so much other stuff, that Jekyll doesn't use.

Gemfile
```diff
@@ -3,7 +3,7 @@ source "https://rubygems.org"
 ruby "2.4.1"
 
 gem 'rake', '~>12.0.0'
-gem 'jekyll', '~> 3.5.1'
+gem 'jekyll', '~> 3.5.2'
```
mlocher: The `Gemfile` doesn't need to specify the latest version`~>` will make sure that the lockfile references the latest patch version automatically.

ryanjwilke: Problem is, running `bundle` will undo the changes to the `Gemfile.lock` file, which is not what you want either.

screenshot 2017-08-24 14 13 09

mlocher: I just checked and according to the net `bundle` is equivalent to `bundle install`. From the docs for `bundle install`:

> If a Gemfile.lock does exist, and you have not updated your Gemfile(5), Bundler will fetch all remote sources, but use the dependencies specified in the Gemfile.lock instead of resolving dependencies.

Which imo indicates that having `~> 3.5.1` in the `Gemfile` should be fine, as long as the `Gemfile.lock` specifies the version we actually want to install (in this case `3.5.2`). 

We've been using it this way for a while now, and have never run into this problem.

mlocher: However, it also doesn't matter to make that change, it's just not required to upgrade to 3.5.2 :)

PR #971, opened 2017-08-25
Modify inline code block colors

kellyjandrews:
The current color scheme of inline code blocks creates large patches of black that cause readability issues. 
screen shot 2017-08-24 at 9 33 11 pm

I updated the color scheme to represent the more common scheme on Github with a light grey background and dark text.  
screen shot 2017-08-24 at 9 32 29 pm

### DIFF
--- a/_sass/docs/_page.scss
+++ b/_sass/docs/_page.scss
@@ -333,22 +333,22 @@ div#integrations {
 }
 
 .alert-block {
-  background-color: $red-lightest;
-  border-color: $red-lightest;
+  background-color: $yellow-lightest;
+  border-color: $yellow-lightest;
 
   > *:first-child::before {
     content: '\f071';
-    color: $red-darkest;
+    color: $yellow-darkest;
   }
 
   a {
-    color: $red-darkest;
+    color: $yellow-darkest;
   }
 
   code {
-    color: $red-darkest;
-    background-color: $red-light;
-    border-color: $red;
+    color: $yellow-darkest;
+    background-color: $yellow-light;
+    border-color: $yellow;
   }
 }
 
--- a/_sass/docs/_page.scss
+++ b/_sass/docs/_page.scss
@@ -376,10 +376,6 @@ div#integrations {
     color: $_cDarkgrey;
   }
 
-  code {
-    background-color: $_cDarkgrey;
-  }
-
   .rating-nero-value {
     margin-right: 2em;
   }
--- a/_sass/docs/_page.scss
+++ b/_sass/docs/_page.scss
@@ -324,52 +324,73 @@ div#integrations {
     margin: 0;
     text-decoration: underline;
   }
+
+  code {
+    color: $blue-darkest;
+    background-color: $blue-light;
+    border-color: $blue;
+  }
 }
 
 .alert-block {
-  background-color: #ffc6cb;
-  border-color: #ffc6cb;
+  background-color: $red-lightest;
+  border-color: $red-lightest;
 
   > *:first-child::before {
     content: '\f071';
-    color: $_cDarkred;
+    color: $red-darkest;
   }
 
   a {
-    color: $_cDarkred;
+    color: $red-darkest;
   }
 
   code {
-    background-color: $_cDarkred;
+    color: $red-darkest;
+    background-color: $red-light;
+    border-color: $red;
   }
 }
 
 .info-block {
-  background-color: #9ed8f5;
-  border-color: #9ed8f5;
+  background-color: $blue-lightest;
+  border-color: $blue-lightest;
 
   > *:first-child::before {
     content: '\f05a';
-    color: $_cDarkblue;
+    color: $blue-darkest;
   }
 
   a {
-    color: $_cDarkblue;
+    color: $blue-darkest;
   }
+
+  code {
+    color: $blue-darkest;
+    background-color: $blue-light;
+    border-color: $blue;
+  }
+
 }
 
 .feedback-block {
   padding-top: .1em;
-  background-color: #f2f5f7;
-  border-color: #f2f5f7;
+  background-color: $gray-lightest;
+  border-color: $gray-lightest;
 
   > *:first-child::before {
     content: '\f059';
-    color: $_cDarkgrey;
+    color: $gray-darkest;
   }
 
   a {
-    color: $_cDarkgrey;
+    color: $gray-darkest;
+  }
+
+  code {
+    color: $gray-darkest;
+    background-color: $gray-light;
+    border-color: $gray;
   }
 
   .rating-nero-value {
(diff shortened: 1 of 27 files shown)


mlocher:
@ryanjwilke What's your take on this? And should we update code blocks at the same time to keep aligned with how we display inline code?

(If we adapt code blocks, we'd likely need to switch to a different syntax highlighting schema, which isn't hard at all, but sth we shouldn't forget.)

ryanjwilke:
@mlocher We should have some styles available in Shipyard already for this. I'll have a quick look.

_sass/rig/_global.scss
```diff
@@ -281,6 +281,11 @@ code {
   transform: translateY(-1px);
 }
 
+p > code, li > code{
```
ryanjwilke: Ideally, it's good to avoid this kind of specificity in the CSS. It becomes hard to overwrite it later if needed. It's also pretty slow in terms of how quickly the browser is able to render the style.

_sass/docs/_page.scss
```diff
@@ -376,10 +376,6 @@ div#integrations {
     color: $_cDarkgrey;
   }
 
-  code {
```
ryanjwilke: I think we should be able to keep this, and just to an `@extend .code` to pull in the styles from shipyard, but I'll have to confirm locally.

_sass/rig/_global.scss
kellyjandrews: What would be a better approach?  Another way to do it would be to make `code` this by default, and override in the other areas - would that make more sense?

_sass/docs/_page.scss
kellyjandrews: I removed this to remove the dark background - if we want to keep the darker background in this area, we can leave the background and change the color of the text.

_sass/rig/_global.scss
kellyjandrews: I think what I was running into was there are a few spot it seemed to need to override the default, so I instead went this route.

ryanjwilke:
@kellyjandrews @mlocher We should be all set in the latest update. We're now making use of the styles from Shipyard.

screenshot 2017-08-28 11 05 01

_sass/rig/_global.scss
```diff
@@ -281,6 +281,19 @@ code {
   transform: translateY(-1px);
 }
 
+p > code,
```
ryanjwilke: @kellyjandrews This should be fine for now. We'll leave it like this for now. Ideally, we want to avoid specificity, but in this case it seems like this is our easiest bet given the structure.

mlocher:
@ryanjwilke What about code blocks (non inline), do we want to adapt those (and switch them to the styles from `shipyard` at the same time?

Imo having light inline code and dark code blocks looks a bit strange (to me)

kellyjandrews:
I was more concerned with readability, which the blocks didn't have any issues really. Since this is non-critical, I'll look at flipping the blocks as well, and try to learn shipyard a bit more too.

ryanjwilke:
@kellyjandrews I just documented the styles better in shipyard here and made some new changes so I'll bump us to the next version for that: https://shipyard-framework.github.io/shipyard/components/code

ryanjwilke:
I'd say this is good to ship now, introducing the styles for `.code-block` should be done in a separate PR as there are going to be some more styles to account for in regards to syntax highlighting, etc...

mlocher:
Did somebody check how inline code in warning/info blocks look like with the new styles? Below is a screenshot of the old version.

ryanjwilke:
@mlocher I only checked the info block; not the warning block. Do we already have a page that has an example of that style?

mlocher:
@ryanjwilke Yeah, see https://documentation.codeship.com/pro/continuous-deployment/google-cloud/ (should have linked that right away with the screenshot)

kellyjandrews:
Does not work as is.

Also, @ryanjwilke - I'm not a fan of the purple text here.  Would using the dark grey text be invalidating something else?

screen shot 2017-08-29 at 8 15 27 am

kellyjandrews:
I can dig into this later today.

ryanjwilke:
screenshot 2017-08-29 15 51 30

@mlocher I fixed the `alert` issue by referencing the styles inside of [Shipyard](https://shipyard-framework.github.io/shipyard/components/code), and documenting those better. The screenshot above is an example of how that looks now.

@kellyjandrews I would prefer to keep the purple text for now as that's consistent with how we're using it inside mothership already, but at some point in the future it may need to be toned down. Let's give it a try for now for the sake of consistency.

mlocher:
@ryanjwilke What happened to the icon, why did the color change, and as the structure of the block is now new, we'd need to update a couple other blocks as well.

Is it at all possible to only fix the code blocks without (for now) touching the HTML, and keep the icons intact?

ryanjwilke:
@mlocher I think we can add some icons back in the future, but for the time being, we're going to need some in-between states in order to bring the docs more inline with the mothership styles. We'll have to tackle it iteratively, instead of all-at-once (which could take weeks/months).

That said, I did fix any other occurrences of `.alert-block` and switched them to `.alert-warning` which is what's used in Shipyard.

Also, to your point about the `<p>` tags. Jekyll handles that just fine, and that class is needed in order to remove the margins on the paragraph tag which is how it's working by default. This is something I can cleanup more in the future by using something like `{% alert :warning %}`...

kellyjandrews:
I think this is fine. It's a step in the right direction, and we can make iterative changes. 

I didn't mean for this PR to be a burden on anyone, only trying to improve the readability a bit.

mlocher:
I'm still 👎  on the change for the warning blocks. Reason being, that we have different blocks in there (warning & info, plus the feedback block at least), that will now look differently and will have to be instantiated differently.

This is not an improvement for me and also not a step in the right direction.

That said, some of the code changes in this PR actually "fixed" the issue with code blocks in those differently colored blocks, see the following screenshot.



The article from that screenshot is available at https://documentation.codeship.com/pro/quickstart/getting-started-part-three/

As such I'm in favor of reverting the changes to the alert block, making sure all those blocks behave similar to the `info` block and then merging this PR.

Next steps would be to create helpers to generate those blocks, which would then allow us to make changes to their markup more easily and get those changes in for all the blocks we're using right now.

mlocher:
I'll admit, I had the commits prepared. Below are screenshots of the different blocks

Inline and block code blocks


Info block


Warning block


Feedback block


A staging build is currently running at https://app.codeship.com/projects/102044/builds/20154b18-5dbf-4659-8186-48b821157244 and I'll post a link to the deployed site once it's finished.

mlocher:
Staging build is deployed at https://documentation.codeship.com/staging/inline-code/pro/continuous-deployment/google-cloud/

mlocher:
After jumping on a call with @ryanjwilke we decided to update the colors for the `warning-block` to yellow. See the following screenshot

kellyjandrews:
Looks good

ethangj:
✅

PR #973, opened 2017-08-25
Env var update

ethangj:


_pro/builds-and-configuration/environment-variables.md
```diff
@@ -122,18 +122,17 @@ The environment variables Codeship populates are:
 
 ```
 CI_COMMITTER_EMAIL
```
mlocher: What about sorting those variables alphabetically? (And same with the Basic variables.)

ethangj:
@mlocher good idea, updated and merging

PR #978, opened 2017-08-28
Fix broken URL

moritzplassnig:


moritzplassnig:
@ethangj Changed the other IBM article too (same issue with the URLs)

PR #981, opened 2017-08-29
add more info about cloud 66 integration

kamerblauwlicht:


moritzplassnig:
@ethangj @mlocher Looks good to me

_basic/continuous-deployment/deployment-with-cloud66.md
```diff
@@ -19,5 +19,19 @@ Integrating [Cloud 66](http://www.cloud66.com/) with Codeship is as simple as co
 Once you've deployed your stack with _Cloud 66_, you'll see a **Redeployment Hook** URL on your **Stack Information** page. To trigger a new deployment via Codeship, simply add a **Script** deployment to the project and add the following command:
 
 ```bash
-curl -X POST -d "" YOUR_STACK_REDEPLOYMENT_URL
+curl -X POST -d "" https://hooks.cloud66.com/stacks/redeploy/xxxx/yyyy
```
mlocher: are there better placeholders for `xxxx` and `yyyy` available, e.g. `app` and `environment`? (It's been a while since I last worked with Cloud66 and I'm not sure about their terminology any more.

With the current URL we are very likely to receive support requests because it isn't obvious enough that those two are placeholders and need to be replaced with the actual values :(

PR #990, opened 2017-09-05
CLI common issues

ethangj:


_pro/builds-and-configuration/cli.md
```diff
@@ -150,3 +150,11 @@ docker exec CONTAINERID
 ```
 
 Note that you are running your containers, looking up the container ID and then connecting to the running container using the container ID.
+
+## Common Issues
+
+There are several common issues that you may experience when using the Jet CLI:
+
+- `No AES key provided` - this error occurs when the Jet CLI can not find an AES key during encryption or decryption. Note that the key file can be found in your _Project Settings_, must be in the same directory you are running `jet encrypt` from and must be named `jet encrypt`.
```
mlocher: keyfile must be named `codeship.aes` not `jet encrypt` :)

PR #1007, opened 2017-09-14
Update submodules documentation

joesiewert:


mlocher:
@ethangj can you take a look and review this?

joesiewert:
Should already be linked on step 1 or is that not the same article?

PR #1014, opened 2017-09-18
Added meta titles

moritzplassnig:


moritzplassnig:
@ethangj Added another commit. Good to know imo.

PR #1022, opened 2017-09-21
added google sitelinks searchbar

moritzplassnig:


mlocher:
I didn't look into this too much, but I guess that the placeholder given in `target` needs to match the name provided in `query-input`.

```
"target": "https://documentation.codeship.com/#stq={searchbox_target}&stp=1",
"query-input": "required name=search_term_string"
```

PR #1029, opened 2017-09-26
Percy fix

ethangj:


_general/integrations/percy.md
```diff
@@ -19,6 +19,9 @@ redirect_from:
 * include a table of contents
 {:toc}
 
+<div class="warning-block">
+Because we don't have a pull request identifier environment variable, Percy integration is limited in it's ability to generate screenshot comparisons over time.</div>
```
joesiewert: `the Percy integration`

joesiewert: `it's` to `its` in this case I think

joesiewert: IIRC `</div>` needs to be on a separate line to render correctly

ethangj:
@joesiewert Updated

PR #1033, opened 2017-09-28
Dockercfg instructions

ethangj:


_pro/builds-and-configuration/image-registries.md
```diff
@@ -69,9 +69,31 @@ This encryption happens with our [local CLI tool]({{ site.baseurl }}{% link _pro
...
-**Note that if you are using Mac OSX**, the newer versions of Docker have changed to store credentials in the OSX keychain rather than in a configuration file.
+If you are using Apple macOS, the newer versions of Docker have changed to store credentials in the macOS keychain rather than in a configuration file.
 
-To get the appropriate authentication file on OSX, you will need to remove the `credsStore` line from Docker's `config.json` to disable Keychain storage, re-run `docker login` and then use the values it then generates in your updated `dockercfg` as shown above.
+We have created an easy workaround to help you get the appropriate authentication file created on macOS:
+
+- First, add the following credentials, as environment variables, to a file named `credentials.env`:
+
+```bash
+Add the required credentials (username, password and an optional registry URL) to a file on your computer, e.g. called credentials.env.
```
mlocher: I would add backticks around `credentials.env` (otherwise the trailing dot might be misleading)

mlocher: Also that sentence should not be inside the code block, only the three lines starting with `DOCKER_`

_pro/builds-and-configuration/image-registries.md
```diff
@@ -69,9 +69,31 @@ This encryption happens with our [local CLI tool]({{ site.baseurl }}{% link _pro
...
+
+```bash
+docker run -it --rm \
+	--env-file=credentials.env \
+	-v "$(pwd):/opt/data/" \
+	-v "/var/run/docker/sock:/var/run/docker.sock" \
+	codeship/dockercfg-generator /opt/data/dockercfg
+```
+
+**Note** that the `DOCKER_REGISTRY` endpoint can be changed to reference a registry other than Dockerhub, such as Quay.io.
```
mlocher: That only works for registries supporting `docker login` though, so I'm pretty sure this wouldn't work with ECR or GCR, which only provide short lived tokens that we'd need to renew.

mlocher: So I would specify, that the endpoint can be changed to support any (privately) hosted registry based on https://docs.docker.com/registry/

ethangj:
@mlocher updated

PR #1034, opened 2017-09-28
Adding feedback form to end block

ethangj:
Using Wufoo account I created, figure we can leave it up a few weeks and see what happens and then optimize.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -36,9 +36,9 @@ layout: default
       <div class="feedback-block">
 				<h2>Need more help?</h2>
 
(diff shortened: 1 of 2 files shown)


mlocher:
Do we want to have both, the rating and the feedback form up there? (And for changes like this a screenshot would make it way easier to review, imo)

ethangj:
screen shot 2017-09-29 at 8 26 43 am
screen shot 2017-09-29 at 8 26 36 am

ethangj:
@mlocher Screenshots added. And yes, my thought was to keep them as eventually we can make this an on-trigger for the downvote if it makes sense.

PR #1036, opened 2017-09-29
Search breadcrumbs

ethangj:
@mlocher The idea here would be to note when we ship this and then measure 30/60/90 days, in GA, organic search traffic to these pages to see if pages per session and bounce rate changes, as well as look as next page path to see if these links are being used.

This will help us figure out if this is an effective way to assist search traffic and therefore we should roll it out across all articles or if we should revisit how this is being done.

Edit: To clarify, these articles were picked because they're our most popular organic search articles and so make the best test.


ethangj:
@mlocher I would long-term, but I think we should verify this approach makes a difference before worrying abot it.

PR #1040, opened 2017-10-03
Raygun docs

ethangj:


_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+---
+
+* include a table of contents
+{:toc}
+
+## About Raygun
+
+[Raygun](https://raygun.com) lets you collect and track errors and deployments for your applications.
+
+By using Raygun you can keep track of error logs and deployment events easier.
```
mlocher: I would combine this with the previous paragraph.

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Codeship Pro
+
+### Setting Your API Key
+
+You will need to add your Raygun API key to your [encrypted environment variables]({{ site.baseurl }}{% link _pro/builds-and-configuration/environment-variables.md %}) that you encrypt and include in your [codeship-services.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/services.md %}).
+
+###  Installing Raygun Dependency
+
+Raygun maintains a list of modules that can be installed as dependencies for a wide variety of languages and frame works. You will want to [visit their documentation](https://raygun.com/docs) and follow the instructions to use the dependency that is right for your application.
```
mlocher: remove the extra space between _frame work_ in the first sentence.

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+Once you have your API key and dependency installed, you will send data via API calls that you can make in your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}):
+
+
+```yaml
+- name: raygun
+  service: app
+  command: raygun.sh
+```
+
+Notice that in this case we are calling a script named `raygun.sh`. Inside this scrip, we could have a Raygun API call similar to:
```
mlocher: _scriptt_ (missing _t_)

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+client.send(new Error(), { 'mykey': 'beta' }, function (response){ });
```
mlocher: Is that a valid shell command? This looks more like Ruby, and please add a lexer to the code block.

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Setting Your API Key
+
+You will need to add your Raygun API key to your to your project's [environment variables]({{ site.baseurl }}{% link _basic/builds-and-configuration/set-environment-variables.md %}).
+
+You can do this by navigating to _Project Settings_ and then clicking on the _Environment_ tab.
+
+###  Installing Raygun Dependency
+
+Raygun maintains a list of modules that can be installed as dependencies for a wide variety of languages and frame works. You will want to [visit their documentation](https://raygun.com/docs) and follow the instructions to use the dependency that is right for your application.
```
mlocher: Is there a more concrete docs page we can link to (same for Pro above).

Also, remove the extra space between _frame_ and _work_.

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+
+This dependency will need to be installed in the Dockerfile that you build via your [setup commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}).
```
mlocher: There is no `Dockerfile` on _Basic_ :)

_general/integrations/raygun.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Sending Data
+
+Once you have your API key and dependency installed, you will send data via API calls that you can make in your [test or deployment commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}):
+
+```
+client.send(new Error(), { 'mykey': 'beta' }, function (response){ });
```
mlocher: Same comments as above, doesn't look like a shell command, and the code block is missing a lexer.

ethangj:
@mlocher Updated

_general/integrations/raygun.md
```diff
@@ -0,0 +1,84 @@
...
+Once you have your API key and dependency installed, you will send data via API calls that you can make in your [codeship-steps.yml file]({{ site.baseurl }}{% link _pro/builds-and-configuration/steps.md %}):
+
+
+```yaml
+- name: raygun
+  service: app
+  command: raygun.sh
+```
+
+Notice that in this case we are calling a script named `raygun.sh`. Inside this script, we could have a Raygun API call similar to:
```
mlocher: Since the script below now is a Ruby script, I would update this paragraph and the step definition above to call `ruby raygun.rb`, or link to the (rather extensive) deployment tracking script available at https://raygun.com/docs/deployments/bash

Basically, what do we want to show with the snippet? How to track a deployment? In that case I would link to the shell deployment script in their documentation, and show to to add it to the container and then call it.

If we want to show how to track more abstract features, we would likely need to come up with a better (shell based) example.

Imo the problem with the Ruby snippet is, that not every Pro container will have the Ruby interpreter available, and the snippet as is right now doesn't show any use case (other than it's possible to track errors, whatever those are exactly).

ethangj:
@mlocher Updated again

PR #1045, opened 2017-10-10
Timeouts and security language

ethangj:


jschley:
@ethangj , what is the tl;dr of what info you would like included in the **How Is Codeship Built?** section? I will fill it in as soon as I get some guidance from you on what you're looking for.

_general/about/vm-and-infrastructure.md
```diff
@@ -53,4 +53,4 @@ All builds on both Codeship Basic and Codeship Pro have 10gb of disk space alloc
 
 On **Codeship Basic**, a build can run for up to 3 hours, although builds will time out if there is no log activity for 10 minutes.
 
-On **Codeship Pro**, a build can run for up to 2 hours, although builds will time out if there is no log activity for 15 minutes.
+On **Codeship Pro**, a build can run for up to 5 hours, although builds will time out if there is no log activity for 15 minutes.
```
dennisnewel: don't we buy spot instances in 3hr blocks? i think the 2hr were added to keep a healthy buffer before the instance potentially gets brutally killed?
the 5 hours is to stop the ones that someone manages to survive longer than the 3hr spot market, without impacting those few big clients who consistently have builds that manage to go past the 3hr mark (as far as i remember from the conversation a couple of months ago)

ethangj: It's difficult to know what to say, but we know a 3 hour limit is not strictly enforced either.

jschley:
@ethangj please review my addition

ethangj:
@jschley Looks good to me, mind approving the PR?

_general/about/vm-and-infrastructure.md
jschley: confirmed we are using 180 minute spot instances which means there are cases where the build cannot run for more than 3 hours. Would recommend leaving the documented limit at 2 hours

PR #1052, opened 2017-10-12
Notes for retrieving screenshots

joesiewert:


_basic/continuous-integration/browser-testing.md
```diff
@@ -124,3 +125,15 @@ export PHANTOMJS_VERSION=YOUR_DESIRED_VERSION
...
+## Screenshots
+
+During your tests you may want to generate screenshots when tests fail. Codeship Basic starts a new build machine for each build and that machine gets terminated as soon as the build finishes. As a result there is not a simple way to save screenshots from failing builds.
+
+However, you can use a [SSH debug session]({{ site.baseurl }}{% link _basic/builds-and-configuration/ssh-access.md %}) to manually run your failing build again which will generate new screenshots. Then from another terminal window you can use `scp` to copy any screenshot files from the debug machine to your local machine for viewing.
+
+Here is an example command. Note that the port and IP address will come from the debug session you start and the file path will depend on where your tests place the screenshots.
+
+```
+scp -P 65504 rof@54.165.213.72:/home/rof/clone/screenshots/my_screenshot.png .
```
alexneigher: 👏

mlocher: I would use placeholders for the port and the IP address, and substitute `/home/rof` with `${HOME}` (and quote the path since we're now using a variable).

joesiewert: I don't think we can use `${HOME}` in this case because the `scp` command is being run from the user's local machine.  I tested it and it returns the home directory for the machine it is run on.  There is probably a workaround, but for this example it is probably overkill.

mlocher: That's true, in that case leave it off, if the remote path doesn't start with a `/`, it's relative to the users home directory by default.

PR #1056, opened 2017-10-17
Google cloud IAM permissions for GCR

ethangj:


_pro/continuous-deployment/deployment-with-kubernetes.md
```diff
@@ -113,6 +113,19 @@ To accomplish this, we provide an `image_tag` declaration that allows you to set
...
+### Permissions
+
+To use Google Cloud with Codeship Pro, your generator service on Codeship will need to authenticate with an account with the correct IAM permissions.
+
+While the permissions may vary and do change, the minimal required permissions are:
+
+- Container Engine Admin
+- Container Engine Cluster Admin
+- Deployment Manager Editor
+- Storage Object Admin
```
mlocher: I guess you have tested this, but do we really need Admin permissions for those? (It sounds like it might be too much, especially in case there is a different set of write permissions available as well)

PR #1057, opened 2017-10-17
update to API "landing page" on documentation

dennisnewel:
considering this the initial draft

LMK if i've missed something


dennisnewel:
we should probably also move it out of the Basic area? not sure how to do that though

PR #1063, opened 2017-10-25
Update published Docker version to 17.10

rheinwein:
[DO NOT MERGE]

https://github.com/codeship/jet-base-ami/pull/105

I couldn't find any other references to the Docker version, so please let me know if I missed something.

### DIFF
--- a/_data/docker.yml
+++ b/_data/docker.yml
@@ -1,1 +1,1 @@
-version: "17.07"
+version: "17.10"


rheinwein:
@ethangj thanks for being prompt -- but this said do not merge :-D We're not on 17.10 in production yet. It probably won't happen until tomorrow so it's probably worth reverting until we're ready

PR #1066, opened 2017-10-25
Note on renaming projects

ethangj:


ethangj:
@mlocher added support contact, but do we have a rename snippet?

mlocher:
It's the "Switch to a different Repository" snippet (using the updated repository name), which then also updates the name of the project (afaik)

PR #1073, opened 2017-11-07
clarify that we indeed support volumes, just not for top level keys

whiteotter:
image

### DIFF
--- a/_pro/builds-and-configuration/services.md
+++ b/_pro/builds-and-configuration/services.md
@@ -261,7 +261,7 @@ The following features available in Docker Compose are not available on Codeship
   * `tmpfs`
   * `ulimits`
   * `volume_driver`
-  * `volumes` (top-level key)
+  * `volumes` ([we do support volumes]({% link _pro/builds-and-configuration/docker-volumes.md %}), just not as a [top-level key](https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes-volume_driver))
   * `privileged`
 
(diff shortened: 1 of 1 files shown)


_pro/builds-and-configuration/services.md
```diff
@@ -261,7 +261,7 @@ The following features available in Docker Compose are not available on Codeship
   * `tmpfs`
   * `ulimits`
   * `volume_driver`
-  * `volumes` (top-level key)
+  * `volumes` ([we do support volumes]({% link _pro/builds-and-configuration/docker-volumes.md %}), just not as a [top-level key](https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes-volume_driver))
```
mlocher: Can we reword this as e.g. 

> `volumes` (Compose v2 top level definition. Service specific volumes are supported, see "[Using Volumes](https://documentation.codeship.com/pro/builds-and-configuration/docker-volumes/)")

Not really happy with my wording either, it's more a suggestion/starting point for further iteration.

ethangj:
@whiteotter @mlocher I'd be in favor of just taking volumes out of the list entirely.

mlocher:
That would work for me as well, if we have a way to monitor how many people try to use them via the top level entry (could be as "easy" as looking for tickets)

ethangj:
We ignore if it's in top level and it won't work as the user anticipates, so hopefully they contact us and/or find the volumes doc anyway

PR #1077, opened 2017-11-09
Use SSH reset button instead

joesiewert:


mlocher:
@ethangj since you pushed the original PR, could you take a look at this as well?

PR #1085, opened 2017-11-18
pro elixir documentation

whiteotter:
How do we feel about sample project being on my own org? Do we want to get that moved to `codeship-library`?

https://trello.com/c/j5oRIsm2/113-elixir-docker-codeship


mlocher:
> How do we feel about sample project being on my own org? Do we want to get that moved to codeship-library?

Imo we should move it to the `codeship-library` organization.

PR #1094, opened 2017-11-23
updating API link to point to v2 instead of v1

dennisnewel:
v2 is getting ready to go out of Beta, so we shouldn't be pointing to v1 anymore


joesiewert:
Should the URL in the header change too?

screen shot 2017-11-23 at 3 52 33 pm

dennisnewel:
@joesiewert the url should still have v2 in it...just to make sure. v1 will still be available for some time

PR #1108, opened 2017-12-05
Fix redirects

mlocher:
Redirects should always end with a slash otherwise Jekyll will create e.g. an `api.html` which correctly redirects, but is never linked to.


dennisnewel:
ohh....had a vague feeling that might be the case, but since it worked locally i wasn't sure

PR #1109, opened 2017-12-06
start quickstart section with list of examples

whiteotter:


whiteotter:
fyi @kellyjandrews

ethangj:
@whiteotter once the build error is fixed, this looks good to me.

PR #1114, opened 2017-12-10
 Add Engine Yard documentation

joesiewert:


mlocher:
I would switch priorities around, move the custom script provided via https://github.com/codeship/scripts/blob/master/deployments/engine_yard.sh to the highest level and only show how to include it in a _Custom Script_ based deployment (i.e. the `curl` command)

And then below a separate heading, show how you can replicate that script directly within the build. 

That way we still offer a kind of _integrated_ deployment, that is easy to configure, plus all the details for somebody who wants complete control over their deployment script.

joesiewert:
Ready for review/merge.

PR #1115, opened 2017-12-11
Network isolation docs update

ethangj:


moritzplassnig:
@ethangj What about mentioning next to "coming soon" that people could join the private beta? It could be a link to your email address. E.g "coming soon - email us if you want to get early access".

_pro/builds-and-configuration/services.md
```diff
@@ -237,14 +237,21 @@ There are several Docker build flags, such as `-w`, that are not executable on C
 
 These flags should instead be implemented as directives in your Services file, as available. For instance, the `-w` instruction can be replaced with the `working_dir` directive applied to any of your services. Most Compose directives not specifically excluded below should function as expected.
 
+## Container networking
+
+We do not support the top-level `networks` directive (see [unavailable features](#unavailable-features)) - but all containers are started on isolated networks, per step, by default.
+
+Containers should be bidirectionally discoverable without requiring any custom setup and should not require custom network creation.
```
mlocher: I personally don't like the _should_ in this sentence. Either we are certain they are discoverable (which Engineering should be able to confirm), or they are not discoverable. I don't think there's a way for the user to isolate containers if they are on the same network (they could still block ports via a firewall, but the container would still be discoverable).

rheinwein:
LGTM

PR #1149, opened 2018-01-16
Fix heading titles for `cleanup` and `validate`

markphelps:
Fix to match the same heading size as the other commands


mlocher:
@ethangj can you take a look at this

PR #1153, opened 2018-01-16
AWS CLI Basic rewrite

ethangj:


_basic/continuous-deployment/deployment-with-awscli.md
```diff
@@ -28,6 +46,18 @@ Please add the following command in the Setup Command section of your test setti
 pip install awscli
 ```
 
-Please visit [https://aws.amazon.com/cli/](https://aws.amazon.com/cli/) for more information including CLI references.
+### Configuring Authentication
+
+Once the CLI is installed, you will need to run the appropriate AWS CLI login commands via your project's [setup commands]({{ site.baseurl }}{% link _basic/quickstart/getting-started.md %}).
```
mlocher: I would not run the login during setup commands, but as part of the deployment script. The reason being that that allows less room for error and makes it very explicit from which point on the AWS CLI is available and configured to access the AWS account.

ethangj: @mlocher updated

PR #1158, opened 2018-01-19
reference single and multi container configurations for AWS EB

drewkitch:


drewkitch:
@ethangj I switched it up from paragraphs to line items, lmk if thats okay

image

PR #1165, opened 2018-01-23
adding a Note section

freshkurt:
adding a Note section to update docs article that this command is available only after updating to Jet CLI 2.2.0+


mlocher:
Nice catch, thx @freshkurt

PR #1168, opened 2018-01-23
Private git scopes

dennisnewel:
Adding screenshots from Kyle for the scopes that's needed for private git server auth tokens


dennisnewel:
can be released if everything's good - incremental update on the docs

PR #1174, opened 2018-01-26
Fix S3 deployment link

mlocher:


mlocher:
It's not only the link formatting that was off, but the source file needs to be restored as well.

drewkitch:
will close this PR in favor of reviving original S3 doc page

PR #1178, opened 2018-01-31
heathcheck clarify

ethangj:
Blocked pending merge of the fix for this.

### DIFF
--- a/_pro/builds-and-configuration/services.md
+++ b/_pro/builds-and-configuration/services.md
@@ -145,15 +145,13 @@ data:
 
 #### Healthchecks
 
-Codeship supports the `healthchecks` directive for healthchecks built into a Docker image. For images that contain a healthcheck, we will poll for availability every 1 second for up to 60 minutes before proceeding.
+Codeship supports the `healthchecks` directive for healthchecks built into a Docker image. For images that contain a healthcheck, we will check for container availability every 1 second for up to 60 minutes before proceeding.
 
 You can find the health polling status in your logs:
 
 ![Healthchecks logs output]({{ site.baseurl }}/images/docker/healthchecks.png)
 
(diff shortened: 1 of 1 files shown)


_pro/builds-and-configuration/services.md
```diff
@@ -145,10 +145,14 @@ data:
 
 #### Healthchecks
 
-Codeship supports the `healthchecks` directive for healthchecks built into a Docker image. For images that contain a healthcheck, we will poll for availability every 1 second for up to 60 minutes before proceeding. You can find the health polling status in your logs:
+Codeship supports the `healthchecks` directive for healthchecks built into a Docker image. For images that contain a healthcheck, we will check for container availability every 1 second for up to 60 minutes before proceeding.
```
drewkitch: >`HEALTHCHECK` directive

to add further emphasis to the directive only being applicable to the Dockerfile

_pro/builds-and-configuration/services.md
```diff
@@ -145,10 +145,14 @@ data:
...
 #### Healthchecks
 
-Codeship supports the `healthchecks` directive for healthchecks built into a Docker image. For images that contain a healthcheck, we will poll for availability every 1 second for up to 60 minutes before proceeding. You can find the health polling status in your logs:
+Codeship supports the `HEALTHCHECK` directive for Docker images. For images that contain a healthcheck defined in the Dockerfile, we will check for container availability every 1 second for up to 60 minutes before proceeding.
+
+You can find the health polling status in your logs:
 
 ![Healthchecks logs output]({{ site.baseurl }}/images/docker/healthchecks.png)
 
+**Note** that Docker will fail a build that makes three unsuccessful attempts to poll for a healthy state. This can be problematic when using options such as `--interval`, which instruct Docker to poll at a different rate than it's default 30 seconds.
```
markphelps: we should probably say `by default`. They can change this by providing `--retries=N` in their HEALTHCHECK directive. We should link to Docker's [docs](https://docs.docker.com/engine/reference/builder/#healthcheck) if we don't already I think

PR #1184, opened 2018-02-01
Ionic documentation

joesiewert:


_basic/languages-frameworks/ionic.md
```diff
@@ -0,0 +1,44 @@
...
+First, if your project is not already setting a current Node version you will need to [install]({{ site.baseurl }}{% link _basic/languages-frameworks/nodejs.md %}#versions-and-setup) at least Node 6:
+
+```
+nvm install 6
+```
+
+Next you can install Ionic with NPM:
+
+```
+npm install -g ionic@latest
```
mlocher: Are we sure we want to recommend the `latest` version? Or ask users to specify a specific version, but not recommend which one? (E.g. by using an environment variable)

joesiewert: Just went with the recommendation directly from the Ionic docs.  I would guess a lot of projects are just including this in `package.json` already if they use it.

PR #1188, opened 2018-02-02
CI env vars as build args

ethangj:
Blocked pending release


_pro/builds-and-configuration/build-arguments.md
```diff
@@ -125,10 +130,46 @@ app:
...
+  build:
+    dockerfile: Dockerfile
+    args:
+      BRANCH: "{{.Branch}}"
+      CI: "{{.CI}}"
+```
+
+The full list of CI/CD-related variables you can use is:
+
+* `ProjectID` (the Codeship defined project ID)
```
mlocher: Since we add this list in multiple places, can we extract it and include it here (so we only need to update a single file, if we make changes)?

ethangj: It's actually a different list, slightly, from the one for push steps - unfortunately.

PR #1212, opened 2018-03-09
[Debugging Builds Via SSH]: Prompted For Password

carlosrodlop:


mlocher:
Hey @carlosrodlop, thanks for the PR.

Would you mind updating the changes to reflect that the SSH client expects the keys in that location, not necessarily the debug feature on Codeship.

Thanks

PR #1214, opened 2018-03-14
On fail docs

ethangj:
Blocked pending build page release.


mlocher:
Are there any other restrictions for `on_fail`, e.g. not on parallel steps, or only a single child step, ...?

ethangj:
Nope, it can be a child, it can have multiple steps and you can have onfail for as many steps as you'd like.

PR #1217, opened 2018-03-19
Frontend updates

joesiewert:
New Basic version needs to go out first.

### DIFF
--- a/_basic/continuous-integration/browser-testing.md
+++ b/_basic/continuous-integration/browser-testing.md
@@ -128,6 +128,21 @@ export PHANTOMJS_VERSION=YOUR_DESIRED_VERSION
 
 [SlimerJS](https://slimerjs.org) 0.9.5 is installed by default and available in the PATH.
 
+To install [SlimerJS 0.10.3](https://slimerjs.org/download.html) install a compatible Firefox version (38.0 to 52.0) and add the following command to your build steps:
+
+```
+npm install slimerjs
+```
+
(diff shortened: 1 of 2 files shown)


_basic/continuous-integration/browser-testing.md
```diff
@@ -122,10 +128,31 @@ export PHANTOMJS_VERSION=YOUR_DESIRED_VERSION
...
+npm install slimerjs
+```
+
+To install [SlimerJS 1.0.0-beta.1](https://slimerjs.org/download.html#betas) install a compatible Firefox version (57.0) and add the following commands to your build steps:
+
+```
+rm -rf $HOME/.slimerjs
+wget https://download.slimerjs.org/releases/1.0.0-beta.1/slimerjs-1.0.0-beta.1.tar.bz2
+tar xaf slimerjs-1.0.0-beta.1.tar.bz2
+cp -r slimerjs-1.0.0-beta.1 $HOME/.slimerjs
```
mlocher: What do you think about extracting this to a script? Since we already have all the required commands?

mlocher:
@joesiewert What do you think about extending the script in `bin/packages.sh` to collect version information for those tools as well, write them to the `data/basic.yml` file and update those articles to reference that file instead?

That way we could (semi) automate this and make it a bit easier to maintain?

_basic/continuous-integration/browser-testing.md
joesiewert: Since this is a beta version I would rather not bother unless there is demand for it. Once this comes out of beta it should be possible to install with npm.

PR #1220, opened 2018-03-20
Adding deployments to robots allow

ethangj:
Not sure if Swifttype will recognize `allow`, Google does but some others don't.

Will try this out, reindex and see if it works.


mlocher:
We could implement `no-index` meta tags on most tag pages, see https://swiftype.com/documentation/site-search/robots_meta_tag

ethangj:
I guess allow is a Google-only instruction then, will close this and keep digging

PR #1240, opened 2018-03-29
instructions to setup Gemfile & Gemfile.lock

drewkitch:


mlocher:
We shouldn't encourage users to write the `Gemfile.lock` by hand, but show them how to use `bundler` from within a running Docker container to generate it. Something similar along the lines of

```
docker run -it --rm -v $(pwd):/app ruby:2.5.0 /bin/sh "cd /app; bundle install"
```

I don't know if this will work out of the box, since I don't know the `Dockerfile` by heart, but that command shows the basic flow

1. Run a one-off Docker container with the Ruby version you need
2. Mount the dev directory via a bind mount
3. Switch to that directory and run `bundle install`, which will generate the `Gemfile.lock` and since it's run within a volume, share it back to the host.

drewkitch:
Makes sense to not encourage writing Gemfile.lock from scratch. And good opportunity to demonstrate generating Gemfile.lock within a Docker context 👍 

https://github.com/codeship/documentation/pull/1251

PR #1242, opened 2018-04-03
Jet update instructions

ethangj:


_pro/jet-cli/installation.md
```diff
@@ -79,6 +79,10 @@ Usage:
 ...
 ```
 
+### Updating
```
mlocher: The macOS snippet includes instructions for updating via homebrew. I think we should remove those as well and replace them with `jet update`

PR #1245, opened 2018-04-04
Update Yarn version

joesiewert:


mlocher:
This might be a good opportunity to pull out the version number for Yarn into a YML file. We can then work on extending the `packages.sh` script to pull version numbers for some of the installed packages (e.g. Yarn, Firefox, Chrome, ...) and write it to that file. 

Since we run that script for each new AMI to update the list of installed packages, this would be an easy change to make sure the versions are accurate as well. What do you think?

PR #1256, opened 2018-04-17
Heroku deploy

dennisnewel:
Updated documentation to match the new deployment template


_basic/continuous-deployment/deployment-to-heroku.md
```diff
@@ -24,27 +24,30 @@ Codeship makes it easy to deploy your application to Heroku using Codeship's int
 ## Setup Heroku Deployment
 
 ### Step 1 - Navigate to Deployment Configuration
-Navigate to your project's deployment configuration page by selecting _Project Settings_ > _Deploy_ on the top right side of the page.
-
-![Project Settings Deployment]({{ site.baseurl }}/images/continuous-deployment/project_configuration.png)
+Navigate to your project's deployment configuration page by selecting _Project Settings_ on the top right side of the page project page, and then the _Deploy_ option in the sub-nav.
```
jondavidjohn: "of the ~~page~~ project page"

jondavidjohn: Not sure I'd use "sub-nav", don't think that means anything to a user?  Maybe "secondary navigation"?

_basic/continuous-deployment/deployment-to-heroku.md
```diff
@@ -24,27 +24,30 @@ Codeship makes it easy to deploy your application to Heroku using Codeship's int
...
 
 ### Step 3 - Heroku
-Select _Heroku_
+Now we're ready to configure your Heroku deployment.
+
+Select the _Heroku_ template from the list of available deploy templates.
 
 ![Select Heroku]({{ site.baseurl }}/images/continuous-deployment/select_heroku.png)
 
+Note that we occasionally add/remove deployment templates, so the exact location might change.
```
mlocher: Do we really need this note? I assume that users will be able to identify the template based on the logo/name even if we update the available deployments.

dennisnewel: removed

dennisnewel:
I'll add an info block at the top. The difference is mentioned in the changelogs, but we can mention it here too...i was trying to keep the documentation focused on the current version, and not the past ones, but it makes sense in this case :)

PR #1260, opened 2018-04-20
moving access to other repositories doc page to basic space

drewkitch:
Recent user attempted to implement Codeship Basic suggestions on Codeship Pro project. Resolution is to move to Basic doc space. Redirect added to continue supporting common error link raised on Codeship Basic builds.


drewkitch:
@ethangj will do 👍

PR #1261, opened 2018-04-28
Switch to the Shipyard Note component

mlocher:
We were previously using a custom component for `info` (and similar) blocks, but [Shipyard](https://codeship.github.io/shipyard/) has since added a [Note component](https://codeship.github.io/shipyard/components/notes) that fits our needs.

This PR makes a `csnote` helper available, that optionally takes a single argument detailing which type of note is used:

* `{% csnote %}...{% endcsnote %}` for a standard (grey) note
* `{% csnote info %}...{% endcsnote %}` for a info (blue) note
* `{% csnote success %}...{% endcsnote %}` for a success (green) note
* `{% csnote warning %}...{% endcsnote %}` for a warning (yellow) note

Info, Success and Warning notes are customized with additional icons to make it easier to distinguish those UI elements and not rely on color alone.

Staging deployment is available at https://documentation.codeship.com/staging/blocks-v1.1/

### Success Note

(currently not in use)

### DIFF
--- a/_basic/builds-and-configuration/dependency-cache.md
+++ b/_basic/builds-and-configuration/dependency-cache.md
@@ -11,7 +11,7 @@ tags:
   - speed
   - caching
 categories:
-  - Builds and Configuration  
+  - Builds and Configuration
 redirect_from:
   - /basic/getting-started/dependency-cache/
 ---
@@ -19,9 +19,9 @@ redirect_from:
 * include a table of contents
 {:toc}
 
-<div class="info-block">
+{% csnote info %}
 Note that this article addresses caching on **Codeship Basic**. If you are looking for information on caching with **Codeship Pro**, [click here]({{ site.baseurl }}{% link _pro/builds-and-configuration/caching.md %})
-</div>
+{% endcsnote %}
 
 ## What Is The Dependency Cache?
 
--- a/_plugins/note_tag.rb
+++ b/_plugins/note_tag.rb
@@ -1,0 +1,22 @@
+module Jekyll
+	class NoteTag < Liquid::Block
+		def initialize(tag_name, type, tokens)
+			super
+			@type = type.strip.downcase unless type.empty?
+		end
+
+		def render(context)
+			content = super
+			css_classes = "note-#{@type}" if @type
+
+			out = <<~EOF
+				<div class="note #{css_classes}">
+					#{content}
+				</div>
+			EOF
+			out
+		end
+	end
+end
+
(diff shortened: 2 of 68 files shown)


_basic/builds-and-configuration/dependency-cache.md
```diff
@@ -11,17 +11,17 @@ tags:
   - speed
   - caching
 categories:
-  - Builds and Configuration  
+  - Builds and Configuration
```
mlocher: My editor automatically strips whitespace from the end of lines, in case you're wondering what those changes are all about :)

_plugins/note_tag.rb
```diff
@@ -0,0 +1,22 @@
...
+				<div class="note #{css_classes}">
+					#{content}
+				</div>
+			EOF
+			out
+		end
+	end
+end
+
+Liquid::Template.register_tag('csnote', Jekyll::NoteTag)
```
mlocher: I tried using note as the tag, instead of csnote, but that broke the layout and encapsulated the complete content inside the note. I didn't dig into this in more detail, but I assume that this is either a Liquid or Jekyll reserved keyword/feature/...

ryanjwilke: Hmm, not exactly sure what went wrong there. It'd be easy enough to fix it inside of Shipyard if you like. This is what the code looks like from the styleguide examples:
```
{% note :info %}
  <p markdown="1">
    Ten other people are also viewing tickets for [this flight](#).
    Use the code `Travel_Light_2017` before you book your flight.
  </p>
  <button class="btn btn-sm note-cta">Call to Action</button>
{% endnote %}
```

ryanjwilke: @mlocher Oh, it may have had issues with this name if you tried registering another one in addition to the one that's registered already inside of Shipyard.

mlocher: Ah, that could very well be the case. Could you link me to the tag definition in Shipyard? Maybe we don't even need my implementation here and can use the one from Shipyard directly?

ryanjwilke: Sure thing. Here it is: https://github.com/codeship/shipyard/blob/master/lib/shipyard-framework/jekyll/tags/note_tag.rb

mlocher: Thanks, I'll look at this when I next have time, I'd still like to merge the PR so it doesn't get stale. Updating to an (updated) `{% note %}` tag will be easy enough with search and replace in the future)

PR #1262, opened 2018-04-30
add --exit-code flag to heroku deployment

drewkitch:
without db migration exit code enabled, a failing migration may still result in a "successful" build


mlocher:
Hmm, I somehow thought that Heroku had fixed this and the `--exit-code` flag wasn't required anymore 😢

PR #1263, opened 2018-05-01
Adding new categories for search indexing work

ethangj:


_basic/continuous-deployment/deployment-with-cloud66.md
```diff
@@ -8,7 +8,8 @@ menus:
     title: Cloud66
     weight: 16
 categories:
-  - Continuous Deployment       
+  - Continuous Deployment  
+  - Deployment
```
mlocher: What's the difference between `Continuous Deployment` and `Deployment`? Do we need both, can we drop one? Imo we should drop one, since they seem to overlap mostly.

_basic/services/memcached.md
```diff
@@ -13,6 +13,8 @@ redirect_from:
   - /classic/getting-started/memcached/
 categories:
   - Services  
+  - Configuration
+  - Configuration
```
mlocher: Double `Configuration`

_general/about/getting_started_with_ci.md
```diff
@@ -13,7 +13,9 @@ tags:
   - continuous deployment
 categories:
   - About Codeship  
-  - Continuous Deployment  
+  - Continuous Deployment
```
mlocher: Why is this article called _Getting Started with CI_, but not in the `Continuous Integration` category?

_pro/jet-cli/cleanup.md
```diff
@@ -6,6 +6,7 @@ menus:
     weight: 3
 categories:
   - Jet CLI
+  - CLI
```
mlocher: Similar to the `Deployment` category above, do we need both categories? They seem very similar and imo we should try to get by with only one of those.

mlocher:
Oh, and please run a staging build, so we can take a look at the article pages and category overview pages before merging this.

ethangj:
@mlocher https://documentation.codeship.com/staging/new-categories/

ethangj:
@mlocher fixed 

screen shot 2018-05-03 at 12 25 36 pm

PR #1265, opened 2018-05-03
Fix account typo

joesiewert:


mlocher:
How come those don't show up via the spell checker. Can we run a spell checker over the `.spelling` file and remove all offending whitelisted words?

PR #1267, opened 2018-05-04
Letting category pages have titles

ethangj:
This will likely need some refactoring, but the constraints are:

- The SEO plugin title metadata seems not to be used anywhere, as we have default titles overriding in all cases. So, it's useless and in the case of injecting the title into the header.html it created duplicate <title>s

- Jekly frontmatter is yaml and won't take variables, so the title can't be set on the actual category HTML layout page itself as far as I can figure, has to be one level above in a template

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,7 +10,6 @@ zendesk:
   chat: false
 
 # SEO Tag
-title: Codeship Documentation
 twitter:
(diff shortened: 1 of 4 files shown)


ethangj:
@mlocher no rush but could you circle back on this when you have som time?

PR #1313, opened 2018-07-04
updated with Stride and moved to projects category

dennisnewel:
missed the article as it was under `account` and not `projects` - it's a project level setting and should be there :)


joesiewert:
Looks like the new HTML checker is failing because the new tags don't technically exist on the live site yet. I would go ahead and merge this as is and I may need to look at excluding tag checking.

PR #1351, opened 2018-09-03
Pro Getting Started Guide polish

drewkitch:


drewkitch:
HTML proofer is stumbling upon a very strange response from Instagram:

image

Not a normal response for an Instagram handle that doesn't exist..

joesiewert:
Weird, looks good now. 🤔

PR #1356, opened 2018-09-20
updated documentation that reflects github app changes

dennisnewel:


_general/about/permissions.md
```diff
@@ -39,9 +39,9 @@ When we say permissions, we are  talking about access you give Codeship to your
 
 In terms of access you give Codeship, there are two different types that are in play: repository level permissions and access level permissions.
 
-For Codeship to configure your repo correctly, the account that connects a repo needs to have the necessary permissions to setup a webhook etc on the repository. For this, we expect that account to have `admin` permissions (or master or owner depending on your source control system).
+For Codeship to configure your Bitbucket or Gitlab repository correctly, the account that connects a repository needs to have the necessary permissions to setup a webhook etc. For this, we expect that account to have `admin` permissions (or `master` or `owner` depending on your source control system). As for access permissions, these influence what we're allowed to do on your behalf, on a per-build basis. We need access to clone your repo, as well as report back with the test results, but not full admin permissions.
```
ethangj: I know it was already in place, but we can we say something other than "webhook etc"? That's maybe a bit too informal.

I also don't quite the separation between account permissions and access permissions, where one needs to be an admin but the other doesn't?

_general/about/permissions.md
```diff
@@ -39,9 +39,9 @@ When we say permissions, we are  talking about access you give Codeship to your
...
 
-As for access permissions, these influence what we're allowed to do on your behalf, on a per-build basis. We need access to clone your repo, as well as report back with the test results, but not full admin permissions.
+Github works slightly different, as we require you to install the CodeShip Github App and allow the app access to the repositories you want to use on CodeShip. We expect you to have permission to install the app and configure it. Once the app has been installed, users who with to setup new projects mainly need to have access to the repository (and the app also need to have access).
```
ethangj: Let's specify that Github requires it, not us.

Maybe also note that this provides per-repo control and is good for security.

_general/about/permissions.md
```diff
@@ -51,9 +51,13 @@ As mentioned above, Codeship requires both repository and access level permissio
 
 #### Github
 
-- For setting up a new project, we need the account to have `admin` permissions.
-- For regular access, we require read/write permissions to your private repositories so that we can clone the repos and report back status.
-- Like all providers that integrate with Github, we'd love to request fewer permissions than we do, but as we're currently using GitHub's OAuth integration, we're limited to the [few options GitHub provides](https://developer.github.com/v3/oauth/#scopes) (we're asking for `repo` and `user:email` scopes). We are looking to move to the new [GitHub Integration](https://developer.github.com/early-access/integrations/integrations-vs-oauth-applications/) options, to offer you more granular control, in the near future.
+- For setting up a new project, we need the CodeShip Github App installed on your Github organization, and access to the necessary repositories via that app. We'll help you set things up when you create your first project, and once the app is installed you only need to make sure it has access to the repository you want to use in your new project.
```
ethangj: It says they need the CodeShip app installed but does not tell them how to do so or where to go for that, can we add some more direction?

dennisnewel:
updated the permissions page to work in your suggestions. Tried to make it a bit clearer, though i see what you mean.
Let me know if it's still muddy 😄

PR #1382, opened 2018-10-31
using env vars in command steps

drewkitch:


_pro/common-issues/using-env-vars-in-command-steps.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+...results in:
+
+```
+The branch name is: $CI_BRANCH
+```
+
+---
+<br />
```
joesiewert: Should this tag be here?

drewkitch: The horizontal rule is a bit close to the next line of text. If there's an idiomatic, markdown friendly approach here then I'll definitely defer to it.

joesiewert: :shipit:

PR #1386, opened 2018-11-06
Recovered notification clarification.

mfogwell:
Clarified that 'restart build' will not trigger the recovered notification hook.

-general/projects/notifications.md

"Note: does not trigger when using 'Restart Build' from the console"


joesiewert:
:shipit:

PR #1388, opened 2018-11-13
Manual approval env var

ethangj:


_pro/builds-and-configuration/environment-variables.md
```diff
@@ -159,6 +159,7 @@ CI_PROJECT_ID
 CI_REPO_NAME
 CI_STRING_TIME
 CI_TIMESTAMP
+CI_BUILD_APPROVED
```
joesiewert: Can we keep in alphabetic order?

ethangj: 👍

PR #1421, opened 2019-01-17
Build only on PR changes

dennisnewel:
This PR will include changes related to the Build only on PR feature

### DIFF
--- a/_general/projects/skipping-builds.md
+++ b/_general/projects/skipping-builds.md
@@ -14,6 +14,7 @@ tags:
   - testing
   - skipping
   - skip
+  - trigger
 categories:
   - Projects
   - Testing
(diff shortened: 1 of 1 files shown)


ethangj:
@dennisnewel Fine as is, but what do you think or rewriting it into a "Build triggers" article - retitle, add all the options, skipping is just one section

dennisnewel:
Yeah, should probably have a separate page. At this stage the PR was mainly used to show @petra-sargent around. We discovered that the regex option is behind the same feature flag as build only on PR, so we figured we'd put everything in here eventually

ethangj:
Got it, I don’t even know if Id do a separate page but maybe just convert this into the build triggers page and include skipping as part of that.

dennisnewel:
gotcha. makes sense

petra-sargent:
I am closing this PR.  The documentation for Build only on PR is addressed in this PR: https://github.com/codeship/documentation/pull/1440

PR #1426, opened 2019-01-23
Removing PERCY_PROJECT

ethangj:


joesiewert:
Should `dockercfg.encrypted` be part of this PR?

ethangj:
Looks like its the same other than a line being removed, and it passed so I guess it's fine?

PR #1458, opened 2019-03-04
scope htmlproof work to branches starting with 'htmlproof'

drewkitch:
Will allow for selective sweeps of urls


drewkitch:
chatted with Joe offline, process is better served with a scheduled build trigger

PR #1512, opened 2019-06-04
This change updates the headlines in the Getting Started Guide.

petra-sargent:
*This change updates the headlines in the Getting Started Guid for CodeShip Pro

### DIFF
--- a/_pro/quickstart/getting-started-part-five.md
+++ b/_pro/quickstart/getting-started-part-five.md
@@ -1,5 +1,5 @@
 ---
-title: Codeship Pro Introduction Guide Part 5
+title: CodeShip Pro Introduction Guide Part 5
 tags:
   - docker
   - jet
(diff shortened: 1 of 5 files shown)


_pro/quickstart/getting-started-part-five.md
```diff
@@ -91,20 +91,20 @@ Once the new build runs, we can check our log output and see our cache in action
 
 Caching is a really powerful way to speed your builds up. We also have a great article on optimizing your builds overall, as well as making sure your Dockerfile is designed with caching in mind. [You can read that here.](https://blog.codeship.com/speeding-up-your-docker-based-builds-with-codeship/)
 
-## Tests Per Branch
+### Running tags on the master branch
```
ethangj: I'd make this more generic, as it doesn't only apply to master - it can be any branch.

petra-sargent: How about: Running tags on a branch

ethangj:
Works for me

_pro/quickstart/getting-started-part-five.md
drewkitch: It's not very self evident what we mean when introduce a subject called "running tags on a branch". The fact that the attribute happens to be named `tag` is incidental (and unfortunate). Something along the lines of "whitelisting step commands by branch name" would be a _step_ in the right direction.

PR #1514, opened 2019-06-10
This updates ListBucket to have permission to the bucket, SHIP-1238

petra-sargent:
This updates ListBucket to have permission to the bucket SHIP-1238


ctennis:
This change doesn't fix the original reported issue.  An entirely separate block is needed for ListBucket, and the principal has to be just the bucketname, not bucketname slash star.

PR #1516, opened 2019-06-11
changing Codeship to CodeShip

mfogwell:
Went through all _basic, _pro, _general .md files and changed Codeship to CodeShip. I was sure not to change codeship to CodeShip as that was being used in paths and 'codeship-steps.yml'


petra-sargent:
I've been updating to CodeShip in the Getting Started Guide.  Thanks for covering thi.

PR #1520, opened 2019-06-24
This adds product documentation for the Auto-supercedence feature

petra-sargent:
This adds product documentation for the Auto-supercedence feature.


ethangj:
@petra-sargent Typo in the work configuration, see the failed build.

_general/projects/build_trigger.md
```diff
@@ -111,3 +111,37 @@ To run builds only on certain branches:
 4. Enter either full branch names like "master" for exact matches or a regular expressions like "^release.*$" for a search pattern.
 
 5. Click the _Save Configuration_ button.
+
+## Skipping builds on the same branch
+
+You can automatically skip previous builds on the same branch. For example, you may want to skip previous code changes. This allows you to run builds on the latest commit.
```
ethangj: Probably easier to say that you skip builds that are currently in the queue than "previous builds" -- and you don't really skip the code changes, you just skip running a build on them.

_general/projects/build_trigger.md
```diff
@@ -111,3 +111,37 @@ To run builds only on certain branches:
...
+
+Skipping builds happens under the following  conditions: 
+* a build is in progress or running 
+* another build is queued
+
```
ethangj: These conditions are true, but it's a little tough to understand what this is saying. Maybe it helps to say instead "If a build is currently in the queue and another build comes into the queue after it, once the running build stops the first queued build will be skipped to instead just run the most recent build."

or something like that (but less wordy hopefully)

_general/projects/build_trigger.md
```diff
@@ -111,3 +111,37 @@ To run builds only on certain branches:
...
+For CodeShip Pro, any previously queued build is skipped when a new build is queued. 
+
+![Build is queue]({{ site.baseurl }}/images/general/build_supercendence.png)
+
+For CodeShip Basic, any previously queued build is skipped when the last running build finishes or stopped. 
+
+![Build is queue]({{ site.baseurl }}/images/general/basic_build_supercedence.png)
+
+When a build is in the skipped state you can click the drip down arrow and select either _View on GitHub_ or _Project Settings_.
```
ethangj: this is maybe unnecessary, what is the user trying to do here?

mauromorales: @ethangj those are the actions a user can do with a skipped build

1. go to the sources in their SCM
2. go to the project settings

they are also possible on other build states, I guess what is being emphazised here is that these are the *only* two actions a user can do

_general/projects/build_trigger.md
fiunchinho: I'd also highlight here the reasons why you would want to do that, like saving time and having a faster feedback loop.

petra-sargent: Thanks for the feedback.  I updated the content. Please review again.

_general/projects/build_trigger.md
petra-sargent: Thanks for the feedback. I updated the content. Please review again.

_general/projects/build_trigger.md
petra-sargent: Yes, I wanted to emphasize to the user the only actions you can do when the state is skipped. If you don't think that's necessary then I can take it out.

ethangj: Got it. Maybe say that explicitly? "Unlike a finished build, you can not restart a skipped build - you can only..." etc.

_general/projects/build_trigger.md
ethangj: This is definitely better! A little unclear int he second half. I wonder if we can say:

The auto-supercede feature allows you to automatically skip builds that are in the queue if a newer build on the same branch comes in. One example is when you commit only to realize you have an error and make a new commit to correct that error - with this feature, you could skip the commit containing the error from running a build and jump directly to the more recent commit.

ethangj:
@petra-sargent one other note, we interchangeable say supersede and supercede throughout the doc - can we verify standard use?

_general/projects/build_trigger.md
petra-sargent: That sounds good.  I made some edits and updated - let me know what you think.

mauromorales:
Just to follow up on last comment, the term should be "supersede". I'm attaching an image of how it looks in the front end.

Screen Shot 2019-06-24 at 17 46 44

petra-sargent:
Thanks everyone for your feedback.

PR #1598, opened 2020-03-05
Add note about paid subscription

markbiegel:


_general/account/whitelisting.md
```diff
@@ -16,6 +16,8 @@ menus:
 * include a table of contents
 {:toc}
 
+> Note: The IP Whitelisting feature is only available on paid subscriptions.
```
joesiewert: Let's wrap this in the note callout:

```
{% csnote info %}
The IP whitelisting feature is only available on paid subscriptions.
{% endcsnote %}
```

